PR #8, opened 2024-05-08
Chore/connections table

mattdean-digicatapult:
Ticket: [VR-67](https://digicatapult.atlassian.net/browse/VR-70)

This PR adds the `connection` table which for now is simply the initial data we need:

* `company_name`
* `status`
* usual combination of `id`, `created_at`, `updated_at`

I've added the migration for this along with the db service as a model for the UI. One wrinkle I've had is with the package set to build to `commonjs` I cannot get knex the binary to work with Typescript. It's not a major issue as it only affects migrations and seeds but it's another nail in the knex coffin. We should really look for a replacement that we like.

[VR-67]: https://digicatapult.atlassian.net/browse/VR-67?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/knexfile.js
+++ b/knexfile.js
@@ -1,0 +1,35 @@
+// Update with your config settings.
+
+const pgConfig = {
+  client: 'pg',
+  timezone: 'UTC',
+  connection: {
+    host: process.env.DB_HOST || 'localhost',
+    database: process.env.DB_NAME || 'veritable-ui',
+    user: process.env.DB_USERNAME || 'postgres',
+    password: process.env.DB_PASSWORD || 'postgres',
+    port: process.env.DB_PORT || '5432',
+  },
+  pool: {
+    min: 2,
+    max: 10,
+  },
+  migrations: {
+    tableName: 'migrations',
+  },
+}
+
+module.exports = {
+  test: pgConfig,
+  development: pgConfig,
+  production: {
+    ...pgConfig,
+    connection: {
+      host: process.env.DB_HOST,
+      port: parseInt(process.env.DB_PORT || ''),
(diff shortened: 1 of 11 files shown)


knexfile.js
```diff
@@ -0,0 +1,35 @@
+// Update with your config settings.
+
+const pgConfig = {
+  client: 'pg',
+  timezone: 'UTC',
+  connection: {
+    host: process.env.DB_HOST || 'localhost',
```
n3op2, line 7: Rather than guessing decided to comment. Wonder why we are not using `env.get('DB_PORT')` as in `.ts` files

migrations/20240507145215_connection_table.js
```diff
@@ -0,0 +1,28 @@
...
+    def.datetime('updated_at').notNullable().defaultTo(now())
+  })
+}
+
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+exports.down = async function (knex) {
+  await knex.schema.dropTable('connection')
```
n3op2, line 27: shall we drop enums as well?  `connection_status`?

n3op2 (approved):

knexfile.js
mattdean-digicatapult: Because this is only looked at when running the migrations. When we do that we only want the db ens not the whole set of envs the applications needs. I could write a separate envalid loader for that but it's not really needed

PR #9, opened 2024-05-08
An initial seed file that inserts a few records in [connection] table

n3op2:
# What

- seeding some data for connection table, however number got added 🤔 
- updated knexfile with the default seeds path

Screenshot 2024-05-08 at 11 16 11


README.md
```diff
@@ -16,12 +16,28 @@ And then to building the `tsoa` routes:
 npm run tsoa:build
 ```
 
+Followed by database migrations
+
+```bash
+npm run db:migrate
+# seeding data
+npm run db:seed
```
mattdean-digicatapult, line 24: this isn't part of the migrations. Can we document it in a separate block please

knexfile.js
```diff
@@ -17,6 +17,9 @@ const pgConfig = {
   migrations: {
     tableName: 'migrations',
   },
+  seeds: {
+    directory: './seeds/',
+  },
```
mattdean-digicatapult, line 22: this shouldn't be needed? Does it error without?

seeds/connections-seed.js
```diff
@@ -0,0 +1,14 @@
+// Currently based of statuses, variants, should cover all possible scenarios within table boundries
+
+const company_name = 'Seeded company_name Name'
+const variants  = [{ status: 'pending', company_name }, { status: 'pending', company_name: '_@@$@*@&%£*&@£*$*(@£*$)@£*&' }, { status: 'unverified', company_name: 0 },  { status: 'unverified', company_name },{ status: 'verified_them', company_name }, { status: 'verified_us', company_name }, { status: 'verified_both', company_name }, { status: 'disconnected', company_name }]
```
mattdean-digicatapult, line 4: Looking at this we're only linting  the `src` directory. Can you update the prettier command in `package.json` to something like `prettier -c ./src ./migrations ./seeds ./test` and check it then formats this correctly?

mattdean-digicatapult (changes requested):

knexfile.js
n3op2: no, it can be removed and it has been, at first added into test/ folder but then looked at the migrations so followed default so this is no longer relevant as covered in the knex's docs

mattdean-digicatapult (dismissed):

mattdean-digicatapult (approved):

PR #10, opened 2024-05-08
Chore/connection controller

mattdean-digicatapult:
Ticket: [VR-70](https://digicatapult.atlassian.net/browse/VR-70)

This PR adds the list connection controller and some stubs for the views. It includes the unit tests for the controller itself but currently no integration tests. Now on to the fun:

Firstly this is about as good as I think I can get it. There's a bug https://github.com/mochajs/mocha/issues/5139 which means that test names aren't reported correctly because we're dynamically importing `chai`. To get round this I've got some really evil boiler plate that looks like:

```ts
let expect: Chai.ExpectStatic
before(async () => {
  expect = (await import('chai')).expect
})
```

Now I have tried a tonne of stuff to make this better. I can't seem to set globals that `tsimp` will correctly find and I can't get chai to load synchronously. So this is as good as I can think until the above issue is resolved

[VR-70]: https://digicatapult.atlassian.net/browse/VR-70?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


n3op2 (approved):

src/controllers/connection/__tests__/index.test.ts
```diff
@@ -0,0 +1,34 @@
...
+import sinon from 'sinon'
+
+import { toHTMLString, withMocks } from './helpers'
+
+import { ConnectionController } from '../index.js'
+
+describe('ConnectionController', () => {
+  let expect: Chai.ExpectStatic
+  before(async () => {
+    expect = (await import('chai')).expect
```
n3op2, line 11: "thought will be way uglier implementation. Wonder if it's possible to preload this for all test suites"

src/views/connection.tsx
```diff
@@ -0,0 +1,26 @@
...
+        {connections.map((connection) => (
+          <div>
+            <div>{Html.escapeHtml(connection.company_name)}</div>
+            <div>{connection.status}</div>
+          </div>
+        ))}
+      </Page>
+    )
+  }
+}
```
n3op2, line 26: I guess each page will have their own templates. if I'm not mistaken CreateConnection, UpdatedConnection would go here, just to confirm: something like `notifications` we would create NotificationTemplates and relevant to the notification views/html/elements would be defined there?

n3op2 (approved):

src/controllers/connection/__tests__/index.test.ts
mattdean-digicatapult: I have tried to no end to make this load as a global. I just cannot get `tsimp` or `tsx` to see it. This is the best I've managed and it's ugly. We can simplify this once https://github.com/mochajs/mocha/issues/5139 is fixed

PR #12, opened 2024-05-09
Bug/database resolve

mattdean-digicatapult:
Fixes a bug where on load tsyringe isn't able to resolve the database because it has a constructor argument.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "veritable-ui",
-  "version": "0.1.6",
+  "version": "0.1.7",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "veritable-ui",
-      "version": "0.1.6",
+      "version": "0.1.7",
       "license": "Apache-2.0",
       "dependencies": {
         "@kitajs/html": "^4.1.0",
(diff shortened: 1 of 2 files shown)


src/ioc.ts
```diff
@@ -2,10 +2,12 @@ import { IocContainer } from '@tsoa/runtime'
 import { container } from 'tsyringe'
 
 import { logger, Logger, type ILogger } from './logger'
+import Database from './models/db/index.js'
 
 container.register<ILogger>(Logger, {
   useValue: logger,
 })
+container.register(Database, { useValue: new Database() })
```
n3op2, line 10: would need to pay attention to this if we add a new dep to the class

n3op2 (approved):

src/ioc.ts
mattdean-digicatapult: yes

n3op2, line 10:
```suggestion
container.register<IDatabase>(Database, { useValue: new Database() })
```
not sure if we have an export of typeof Database = IDatabase

mattdean-digicatapult: Not necessary because this is a class. tsyringe can use the class as the specifier

PR #15, opened 2024-05-10
Feature/vr 69

n3op2:
# What

Basic layout elements https://digicatapult.atlassian.net/browse/VR-69:
- SideBar - main nav
- Content 
    - Header - rendering some current info e.g. path, form titles and etc
    - Main - for rendering data and user views
    
## screenshots
Screenshot 2024-05-10 at 08 30 08
Screenshot 2024-05-10 at 08 30 00


public/styles/main.css
```diff
@@ -3,13 +3,62 @@
...
-  justify-items: center;
   font-family: system-ui, sans-serif;
-  font-weight: 500;
-  font-size: 10rem;
+}
+
+body.flex-page {
+  display: flex;
+  flex-direction: row;
+  width: 100vw;
```
mattdean-digicatapult, line 15: you probably don't need `width` as it's a block level element. It should just fill available space. Can you try removing any width 100% or 100vw that are applied to divs and check I'm right. This is somewhat important as it reduces the odds of random scrollbars appearing

src/views/common.tsx
```diff
@@ -9,9 +15,16 @@ export const Page = (props: Html.PropsWithChildren<{ title: string }>): JSX.Elem
         <script src="lib/htmx.org/ext/json-enc.js"></script>
         <link rel="icon" type="image/ico" sizes="48x48" href="/public/images/favicon.ico" />
         <link rel="stylesheet" type="text/css" href="/public/styles/main.css" />
+        <link rel="stylesheet" type="text/css" href="/public/styles/semantic.css" />
```
mattdean-digicatapult, line 18: does this exist?

n3op2: nope, tested something earlier another css for some basics e.g. layout

public/styles/main.css
n3op2: this should be updated, thanks for the flag up

mattdean-digicatapult (approved):

PR #16, opened 2024-05-10
Feature/authentication

mattdean-digicatapult:
Ticket [VR-78](https://digicatapult.atlassian.net/browse/VR-78)

Implements authentication for veritable-ui. This is done using https://github.com/digicatapult/tsoa-oauth-express and a cookie based auth implementation. For dev I've included a preconfigure keycloak with a realm and suitable client (appreciate if reviewers can test this) which is launched in the docker compose.

In terms of implementation it's mostly in a  new model `IDPService` and the `AuthController`. 

[VR-78]: https://digicatapult.atlassian.net/browse/VR-78?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/ioc.ts
```diff
@@ -1,14 +1,6 @@
 import { IocContainer } from '@tsoa/runtime'
 import { container } from 'tsyringe'
 
-import { logger, Logger, type ILogger } from './logger'
```
mattdean-digicatapult, line 4: moved this around as it was causing issues in tests because of order of operations

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "veritable-ui",
-  "version": "0.1.15",
+  "version": "0.2.0",
```
syedhasandigi, line 3: A question on package version number, I can see that in each PR this number is incremented, but by how much? Is there any reading material on this?

dblane-digicatapult: Standard SEMVER.

mattdean-digicatapult: It roughly follows [semver](https://semver.org/) but because we're still at version `0.x.x` we can do breaking changes whenever and it's all a bit meaningless. As we progress I'd expect a minor bump (like here) for major feature implementations and patch bumps for things that are more incremental.

This all becomes a bit more meaningfull once we hit v1 and at that moment we need to maintain the "interface" of the application

src/controllers/__tests__/auth.test.ts
```diff
@@ -0,0 +1,224 @@
...
+    .resolves({ access_token: 'access', refresh_token: 'refresh' }),
+} as unknown as IDPService
+
+const mkRequestMock = () => ({
+  res: {
+    clearCookie: sinon.stub(),
+    cookie: sinon.stub(),
+    redirect: sinon.stub(),
+  },
+  signedCookies: { VERITABLE_NONCE: 'nonce', VERITABLE_REDIRECT: '/example' },
```
n3op2, line 28: like the idea of redirect being a property.

src/controllers/__tests__/helpers.ts
```diff
@@ -2,15 +2,26 @@ import { Readable } from 'node:stream'
 
 import pino from 'pino'
 
-import Counter from '../../models/counter'
-import RootTemplates from '../../views/example'
+import { Env } from '../../env.js'
+import Counter from '../../models/counter.js'
```
n3op2, line 6: Hehe... it's still here, I've removed in my branch! so after this I can go

src/controllers/auth.ts
```diff
@@ -0,0 +1,108 @@
...
+  httpOnly: true,
+  signed: true,
+  secure: true,
+  path: '/auth/redirect',
+}
+
+const tokenCookieOpts: express.CookieOptions = {
+  sameSite: true,
+  maxAge: 24 * 60 * 60 * 1000,
+  httpOnly: true,
```
n3op2, line 28: http? why not https? due to (kind/local)?

src/controllers/auth.ts
```diff
@@ -0,0 +1,108 @@
...
+      state: nonce,
+      scope: 'openid',
+    }).toString()
+
+    this.logger.debug('login redirect to %s', redirect)
+    res.redirect(302, redirect.toString())
+  }
+
+  @Get('/redirect')
+  @SuccessResponse(302, 'Redirect')
```
n3op2, line 83: got to say a compliment for status code usage, you've been on top as far as I can remember and this is maybe first time I'm saying (hopefully)

src/env.ts
```diff
@@ -1,6 +1,20 @@
 import * as envalid from 'envalid'
 import { singleton } from 'tsyringe'
 
+const strArrayValidator = envalid.makeValidator((input) => {
+  const arr = input
+    .split(',')
+    .map((s) => s.trim())
+    .filter((s) => !!s)
```
n3op2, line 8: mmm... not sure but my brain wants to tell me something about `bang bang` operator, but can not remember from top oh my head under which circumstances it could not work as expected, but knowing that we have tests and etc pretty sure it will get flagged.

src/errors.ts
```diff
@@ -0,0 +1,23 @@
+export abstract class HttpError extends Error {
+  public abstract get code(): number
+}
+
+export class ForbiddenError extends HttpError {
```
n3op2, line 5: Yay! was just mentioning to the @syedhasandigi about the Error class. I guess if you have not updated maynbe this is something you could use and/or add. I think it could be a good exercise after the story you are working on just to get a better grip on .ts

CC @syedhasandigi

src/server.ts
```diff
@@ -28,16 +35,59 @@ export default async (): Promise<Express> => {
...
 
   app.use('/public', express.static('public'))
   app.use('/lib/htmx.org', express.static('node_modules/htmx.org/dist'))
   app.get('/api-docs', (_req, res) => res.json(swaggerJson))
   app.use('/swagger', serve, setup(undefined, options))
   app.get('/', (_, res) => res.sendStatus(404))
 
+  RegisterRoutes(app)
+
+  app.use(function errorHandler(
```
n3op2, line 49: oh... sound like another story for practise (create error handler)

n3op2 (approved):

src/controllers/auth.ts
mattdean-digicatapult: I think you're misunderstanding what this does https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie#httponly

src/env.ts
mattdean-digicatapult: we're guarenteed to have a string at this point so `!!` will remove falsey strings, notable `''`

syedhasandigi (approved):

PR #21, opened 2024-05-13
Feature/vr 71

n3op2:
# What

Adding page UI elements in style as per figma -> https://www.figma.com/design/nTyayQwsil6Jfy85eVm30E/Veritable-Draft-May-2024?node-id=147-1928&m=dev
- Side Bar - main navigation / main menu
- Content elements along with the header
- Mobile version adjustments for UX

To be fair I think screenshot speak for themselves, please refer below:
Screenshot 2024-05-13 at 11 48 03
Screenshot 2024-05-13 at 11 47 59
Screenshot 2024-05-13 at 11 47 54
Screenshot 2024-05-13 at 11 47 45
Screenshot 2024-05-13 at 11 47 41


public/styles/main.css
```diff
@@ -1,18 +1,58 @@
+/* GENERICS - project wide ------------------------ 
+- TO Maybe: once enough generics - abstract to it's own file?
```
mattdean-digicatapult, line 2: yes we should have these in a separate file. Normally I'd call this a "reset" so maybe `reset.css`?

public/styles/main.css
```diff
@@ -1,18 +1,58 @@
...
+/* END OF GENERICS ------------------- */
+
 body {
   margin: 0;
   padding: 0;
   display: grid;
-  font-family: system-ui, sans-serif;
+  background-color: #f4f5fb;
+  color: #45494c;
+  font-family: system-ui, Roboto;
```
mattdean-digicatapult, line 49: should this not be `Roboto` first? We'll always have `system-ui` on all systems

public/styles/main.css
```diff
@@ -23,35 +63,129 @@ body.flex-page {
...
   display: flex;
   flex-direction: column;
-  min-height: 100vh;
 }
 
 .content.header {
+  display: flex;
+  flex-direction: column;
+  align-items: flex-start;
+  width: 95vw; /* due to column flex-grow: grows height so to void double wrap of els e.g. col/row */
```
mattdean-digicatapult, line 137: If this is calculated from the side bar width can we add a custom variable to the body element (`--sidebar-width`) and then do a calc here?

public/styles/main.css
```diff
@@ -23,35 +63,129 @@ body.flex-page {
...
+}
+
+.header.nav {
   display: flex;
   flex-direction: row;
-  align-items: center;
-  min-height: 5vh;
+  border-top: #c9cace 0.1px solid;
+
+  width: 100%;
```
mattdean-digicatapult, line 152: do we need this width 100%?

public/styles/main.css
n3op2: This has been updated, just waiting on checks.

public/styles/main.css
n3op2: as per slack conversation I've added roboto and fallback - system

public/styles/main.css
n3op2: yes, due to mobile version not having yet a content element, otherwise it leave a little default background color

public/styles/reset.css
```diff
@@ -0,0 +1,57 @@
...
+:root {
+  --mobile-header-bg: transparent;
+  --side-bar-width: 5vw;
+  --desktop-content-width: calc(100vw - 5vw);
+  --accent-color: #5670f1;
+  --secondary-color: #FFCC91;
+  --sub-text-color: #c9cace;
+  --text-color: #45494c;
+  --bg-color: #f4f5fb;
+}
```
mattdean-digicatapult, line 24: this is application specific and shouldn't be part of a reset

mattdean-digicatapult (approved):

mattdean-digicatapult (approved):

PR #22, opened 2024-05-13
Feature/vr 56 Verify Company Name, Check in to see progress

syedhasandigi:
---
name: Feature
about: This feature will allow the UI to verify a company house number against the company house website
---

## Summary
This is a PR to ticket [https://digicatapult.atlassian.net/browse/VR-56](url)

This feature provides the user with a form that allows them to verify a company by entering the company number into a form which will then dynamically populate a text box on the UI that contains either a valid company or an error response. This feature is a precursor to submitting a connection request to a company on the veritable UI

## Motivation

By adding this feature the user will know if the company does in fact exist and can confirm the address and status of the company before attempting to connect with them

N/A

## To do
[x] Write tests for model
[x] Create model
[] Write tests for htmx Controllers
[x] Create controllers
[x] Build views

## To test
`git checkout feature/vr-56/verify-company-name`
`npm run test`

## To run
`git checkout feature/vr-56/verify-company-name`
`npm run dev` - note that currently there are some bugs that I have not finished working on that will not allow the server to start, will be fixing these in my next commit to the branch

## Feedback
I'm looking for feedback on my approach to the problem, 
in what ways could this be improved, 
have I followed best practices?


src/controllers/newConnection/index.ts
```diff
@@ -0,0 +1,48 @@
+import { Get, Produces, Route, SuccessResponse } from 'tsoa'
+import { inject, injectable, singleton } from 'tsyringe'
+
+import { Logger, type ILogger } from '../../logger.js'
+import CompantHouseEntity from '../../models/companyHouseEntity.js'
+import NewConnectionTemplates from '../../views/newConnection.js'
```
n3op2, line 6: since you mentioned in the stand up, I follow '-' when it comes to naming files e.g. new-connection.ext and if it would be an instance I would go with the Capital e.g. NewConnection.ext. I guess everybody has got their own preferences here.

I wanted to point out that we already have a view for `Connection` could we put a form in there maybe to keep connection stuff together? Not suggesting to change or anything this would not block me from approving a PR.

src/models/__tests__/helpers/mockCompanyHouse.ts
```diff
@@ -0,0 +1,64 @@
+import { container } from 'tsyringe'
```
n3op2, line 1: nice for doing a mock, looks good.

src/models/__tests__/helpers/mockCompanyHouse.ts
```diff
@@ -0,0 +1,64 @@
...
+    address_line_1: '51 Mornington Crescent',
+    locality: 'Hounslow',
+    postal_code: 'TW5 9ST',
+    country: 'United Kingdom',
+  },
+  company_status: 'dissolved',
+  registered_office_is_in_dispute: false,
+  company_name: 'SMH IOT SOLUTIONS LTD',
+  company_number: '10592650',
+}
```
n3op2, line 22: maybe create something like `fixtures/` folder for this kinda stuff but again, is just 10 lines so I would not worry too much

src/models/companyHouseEntity.ts
```diff
@@ -0,0 +1,75 @@
...
+
+    const response = await fetch(url.toString(), {
+      method: 'GET',
+      headers: new Headers({
+        Authorization: 'Basic ' + process.env.COMPANY_PROFILE_API_KEY,
+      }),
+    })
+
+    if (!response.ok) {
+      throw new Error(`Error calling CompanyHouse API`)
```
n3op2, line 59: once we have more Errors we can abstract them like we do elsewhere -> https://github.com/digicatapult/sqnc-hyproof-api/blob/main/src/lib/error-handler/index.ts

src/views/newConnection.tsx
```diff
@@ -0,0 +1,56 @@
...
+  public formPage = (company: CompanyProfile, validated: boolean) => {
+    return (
+      <Page title="Veritable - New Connection">
+        <this.CompanyNumberInput validated={false} />
+        <this.CompanyTextBox company={company} />
+      </Page>
+    )
+  }
+
+  public CompanyNumberInput = ({ validated }: { validated: boolean }) => {
```
n3op2, line 20: Let me know once it's building would be keen on trying this

src/views/newConnection.tsx
```diff
@@ -0,0 +1,56 @@
...
+          <p>{company.registered_office_address.locality}</p>
+          <p>{company.registered_office_address.po_box}</p>
+          <p>{company.registered_office_address.postal_code}</p>
+          <p>{company.registered_office_address.premises}</p>
+          <p>{company.registered_office_address.region}</p>
+          <p>{company.company_status}</p>
+      </div>
+    )
+  }
+}
```
n3op2, line 56: not sure if UI was part of the story, so in this instance anything will do for as long as we can validate it works. even if it's `JSON.stringify(response)` in plain text also we can verify using the Dev tools "Networking Tab"

src/controllers/newConnection/index.ts
dblane-digicatapult: I'm not sure we have an existing convention on this. @mattdean-digicatapult ?

syedhasandigi: So my reasoning behind creating a newConnections directory for the routes is that if we put them all in the Connections route it would get cluttered, as for each element in HTMX you may need multiple routes/endpoints to perform logic. So my opinion, which probably is flawed, is to put each "VIEW/FORM" in its own organised directory.

As for naming conventions, I looked at some repositories like veritable-cloudagent, and saw that there are some inconsistencies, I'm happy to change to whatever convention the team agrees upon.

src/models/__tests__/helpers/mockCompanyHouse.ts
syedhasandigi: Noted, I will find an example and follow, I'm sure as we go along there will be many more fixtures.

src/env.ts
```diff
@@ -1,3 +1,4 @@
+import 'dotenv/config'
```
syedhasandigi, line 1: Not sure why this import is needed

src/controllers/connection/newConnection.ts
```diff
@@ -0,0 +1,75 @@
...
+import { inject, injectable, singleton } from 'tsyringe'
+
+import { Logger, type ILogger } from '../../logger.js'
+import CompantHouseEntity from '../../models/companyHouseEntity.js'
+import NewConnectionTemplates from '../../views/newConnection.js'
+import { HTML, HTMLController } from '../HTMLController.js'
+
+@singleton()
+@injectable()
+@Route('/new-connection')
```
mattdean-digicatapult, line 11: probably should be `/connection/create` or `/connection/new`

src/controllers/connection/newConnection.ts
```diff
@@ -0,0 +1,75 @@
...
+@Produces('text/html')
+export class NewConnectionController extends HTMLController {
+  constructor(
+    private companyHouseEntity: CompantHouseEntity,
+    private newConnection: NewConnectionTemplates,
+    @inject(Logger) private logger: ILogger
+  ) {
+    super()
+    this.logger = logger.child({ controller: '/' })
```
mattdean-digicatapult, line 20: this should match the path of the controller so at the moment `/new-connection`

src/controllers/connection/newConnection.ts
```diff
@@ -0,0 +1,75 @@
...
+    this.logger.debug('new connection page requested')
+    return this.html(this.newConnection.formPage())
+  }
+
+  /**
+   * Returns a company from a validated company number
+   */
+  @SuccessResponse(200)
+  @Get('/verify-company')
+  public async verifyCompanyForm(@Query() companyNumber: string): Promise<HTML> {
```
mattdean-digicatapult, line 39: as discussed we should probably make this replace the whole form. It'll just make further validation easier

src/env.ts
mattdean-digicatapult: it's a really annoying mutation side-affect. Basically it populate `process.env` when you call this import

src/env.ts
```diff
@@ -29,6 +30,8 @@ const envConfig = {
   IDP_OIDC_CONFIG_URL: envalid.url({
     devDefault: 'http://localhost:3080/realms/veritable/.well-known/openid-configuration',
   }),
+  COMPANY_HOUSE_API_URL: envalid.str({ default: 'https://api.company-information.service.gov.uk' }),
+  COMPANY_PROFILE_API_KEY: envalid.str({ devDefault: 'APIKEY' }),
```
mattdean-digicatapult, line 34: Assuming this APIKEY doesn't work we should probably leave this unset.

src/models/__tests__/helpers/mockCompanyHouse.ts
```diff
@@ -0,0 +1,64 @@
...
+
+export function companyHouseMock() {
```
mattdean-digicatapult, line 24: can we rename this to `withCompanyHouseMock`

src/models/companyHouseEntity.ts
```diff
@@ -0,0 +1,74 @@
...
+    const url = new URL(route)
+
+    const response = await fetch(url.toString(), {
+      method: 'GET',
+      headers: new Headers({
+        Authorization: this.env.get('COMPANY_PROFILE_API_KEY'),
+      }),
+    })
+    if (!response.ok) {
+      console.log(response)
```
mattdean-digicatapult, line 57: remove console log please. If you feel the logging is helpful replace it with a `logger.debug` call and add the logger as a dependency in the constructor (look at the controllers if you need an example)

mattdean-digicatapult (changes requested):
really well done @syedhasandigi! This is a great first try with just a few small changes needed

package.json
```diff
@@ -46,6 +47,7 @@
     "swagger-ui-express": "^5.0.0",
     "tsoa": "^6.2.1",
     "tsyringe": "^4.8.0",
+    "undici": "^6.16.1",
```
mattdean-digicatapult, line 50: undici is already a dev dependency, you should be able to just remove this line and `npm i`

src/models/companyHouseEntity.ts
mattdean-digicatapult: We already have https://github.com/digicatapult/veritable-ui/blob/main/src/errors.ts. This would be an `InternalError`

mattdean-digicatapult (changes requested):

mattdean-digicatapult (approved):

PR #28, opened 2024-05-17
Fix missing env var

rmlearney-digicatapult:
---
name: Bug report
about: Create a report to help us improve
---

### Prerequisites

- [X] Put an X between the brackets on this line if you have done all of the following:
  - Checked the FAQs for common solutions: <https://github.com/digicatapult/veritable-ui/blob/main/CONTRIBUTING.md/#FAQs>
  - Checked that your issue isn't already filed: <https://github.com/issues?utf8=✓&q=is%3Aissue+user%3Averitable-ui>

### Description

```
 ~/git/veritable-ui/ [main] npm run dev

> veritable-ui@0.3.3 dev
> npm run tsoa:watch & NODE_ENV=dev node --import=tsimp/import --watch src/index.ts


> veritable-ui@0.3.3 tsoa:watch
> node --watch-path=./src ./node_modules/.bin/tsoa -- spec-and-routes

Completed running './node_modules/.bin/tsoa -- spec-and-routes'
================================
 Missing environment variables:
    COMPANY_PROFILE_API_KEY: undefined
================================

 Exiting with error code 1
Failed running 'src/index.ts'
```

### Steps to Reproduce

1. Follow instructions in `readme.md`

**Expected behaviour:**

Starts server

**Actual behaviour:**

Borks with `COMPANY_PROFILE_API_KEY: undefined`

**Reproduces how often:**

100%


mattdean-digicatapult:
I'd rather not have a non-functional env var in the application. There are two improvements I do think could be made here:

1. Instructions in the readme on where to get an api key from and how to use it in dev
2. A switch when running in dev mode that set's up an undici mock so that we always get valid values from companies house. We might even be able to do it with a `--import` so that the logic can only ever apply to the `npm run dev` command

PR #37, opened 2024-05-28
VR-55 - New connection request form, No styling

syedhasandigi:
# Pull Request

## Checklist

## PR Type

- [ ] Feature

## Linked tickets

https://digicatapult.atlassian.net/browse/VR-55

## High level description

Form is complete, styling and submission into database missing

## Detailed description

- [ ] Alice can submit details into a form in a frontend ([Design reference](https://www.figma.com/file/nTyayQwsil6Jfy85eVm30E/Veritable-Draft-May-2024?type=design&node-id=102-5927&mode=design&t=Tn5uxxV2cAm03OCr-0))
    - [x] HTML functionality exist
    - [x] Dynamically move through stages
    - [ ] Styling

- [x]  With form validation that checks a few behaviours:
   - [x] Prevents continuing (on value update + submit)
- [x] Want to check if email format is correct
- [x] After submission of the form, Alice will see a screen that her connection request has been submitted. 
- [ ] The connection email address will be added into a database

## Additional context

While trying to implement email validation by using a zod type, A `no matching model` error is generated

https://github.com/lukeautry/tsoa/issues/1256

The way I've done it feels clunky but it works.


src/controllers/connection/newConnection.ts
```diff
@@ -76,9 +92,69 @@ export class NewConnectionController extends HTMLController {
...
-  @Get('/submit')
-  public async submitCompanyNumber(): Promise<HTML> {
+  @Post('/submit')
+  public async submitCompanyNumber(
+    @Body() body: { companyNumber: string; email: string; formStage: string; submitButton: string }
+  ): Promise<HTML> {
     // do some regex if there is a match on the whole regex return true else return false
-    return this.html(this.newConnection.companyEmptyTextBox({ errorMessage: 'Company does not exist' }))
+
+    try {
```
syedhasandigi, line 101: The clunkyness

src/views/newConnection.tsx
```diff
@@ -1,42 +1,118 @@
...
       </Page>
     )
   }
 
-  public companyNumberInput = ({ targetBox }: { targetBox: JSX.Element }): JSX.Element => {
+  public companyFormInput = (params: {
+    targetBox: CompanyProfileText
+    formStage: FormStage
+    email?: string
+    companyNumber?: string
```
mattdean-digicatapult, line 37: we need to be more careful with these values as they come form the user. Can we please make sure they are escaped to avoid xss

src/views/newConnection.tsx
```diff
@@ -80,10 +156,18 @@ export default class newConnectionTemplates {
...
   public companyEmptyTextBox = ({ errorMessage }: { errorMessage: string }): JSX.Element => {
-    return (
-      <>
-        <p>{Html.escapeHtml(errorMessage)}</p>
-      </>
-    )
+    return <>{Html.escapeHtml(errorMessage)}</>
+  }
+
+  public stepper = (params: { formStage: FormStage }): JSX.Element => {
```
mattdean-digicatapult, line 162: This will definitely need to change but that's fine for now

src/views/newConnection.tsx
```diff
@@ -1,42 +1,118 @@
...
-            maxlength={8}
-            oninput="this.reportValidity()"
-          ></input>
-          <input id="email" placeholder="Connection's Email Address" name="email"></input>
-          <button id="btn-cancel">Cancel</button>
-          <button id="btn-continue">Continue</button>
-          <div id="text-box-target">{targetBox}</div>
+        <form
+          id="new-connection-form"
+          class="new-connection"
```
mattdean-digicatapult, line 46: what does `new-connection` represent? It seems to be on a lot of elements with very different meaning. How bad would it be to remove all style classes?

mattdean-digicatapult (changes requested):

public/styles/main.css
```diff
@@ -124,6 +124,57 @@ a.connections-table.icon {
...
+}
+
+.new-connection button {
+  display: flex;
+  border-color: #abafb1;
+  flex-direction: column;
+  align-items: flex-start;
+  gap: 8px;
+  align-self: stretch;
+}
```
n3op2, line 161: I guess this can be ignored since it's mention in the PR that this is excluding styling.

src/controllers/connection/newConnection.ts
```diff
@@ -44,30 +48,41 @@ export class NewConnectionController extends HTMLController {
...
+          targetBox: {
+            status: 'error',
+            errorMessage: 'Company number format incorrect',
+          },
+          formStage: 'form',
         })
       )
     }
-    let company
+    let company: CompanyProfile
```
n3op2, line 64: wonder if this could be added as a function variable with a default value like `null`

so function would look like

```js
public async verifyCompanyForm(@Query() companyNumber: string, company: CompanyProfile = undefined): Promise<HTML> {
  ...
}
```

This way we could avoid using mutable `let`

src/views/newConnection.tsx
```diff
@@ -1,42 +1,118 @@
...
+              name="submitButton"
+              value="Back"
+              display={showConfirmation ? 'block' : 'none'}
+            />
+            <FormButtonIcon
+              type="submit"
+              name="submitButton"
+              value={`${showForm ? 'Continue' : 'Submit'}`}
+              display={showSuccess ? 'none' : 'block'}
+            />
```
n3op2, line 107: Since it's same type and same name. maybe we could combine into a single one as per below

```js
            <FormButtonIcon
              type="submit"
              name="submitButton"
              value={showConfirmation ? "Back" : showSuccess ? "Continue" : "Submit"}
            />
```
Not sure about the typing might need to double check, but this way we do not need to use "display" additional <FormButtonIcon />

src/views/newConnection.tsx
```diff
@@ -80,10 +156,18 @@ export default class newConnectionTemplates {
...
+    return <>{Html.escapeHtml(errorMessage)}</>
+  }
+
+  public stepper = (params: { formStage: FormStage }): JSX.Element => {
+    if (params.formStage === 'form') {
+      return <div>-----Step 1</div>
+    } else if (params.formStage === 'confirmation') {
+      return <div>----------Step 2</div>
+    } else if (params.formStage === 'success') {
+      return <div>---------------Step 3</div>
```
n3op2, line 168: like the measurement of the step e.g. `-------`

src/controllers/connection/newConnection.ts
```diff
@@ -43,31 +48,28 @@ export class NewConnectionController extends HTMLController {
...
-      /^(((AC|CE|CS|FC|FE|GE|GS|IC|LP|NC|NF|NI|NL|NO|NP|OC|OE|PC|R0|RC|SA|SC|SE|SF|SG|SI|SL|SO|SR|SZ|ZC|\d{2})\d{6})|((IP|SP|RS)[A-Z\d]{6})|(SL\d{5}[\dA]))$/
-    if (!regex.test(`${companyNumber}`)) {
-      return this.html(
-        this.newConnection.companyNumberInput({
-          targetBox: await this.newConnection.companyEmptyTextBox({ errorMessage: 'Company number format incorrect' }),
-        })
-      )
-    }
-    let company
+  public async verifyCompanyForm(@Query() companyNumber: COMPANY_NUMBER, company?: CompanyProfile): Promise<HTML> {
```
syedhasandigi, line 51: Adding the company here with a zod type generates this error https://github.com/digicatapult/veritable-ui/actions/runs/9286851097/job/25554482149#:~:text=There%20was%20a,in%20%27NewConnectionController.verifyCompanyForm%27

src/views/newConnection.tsx
syedhasandigi: So the two `<FormButtonIcon>` will be shown at the same time in some of the views

src/controllers/connection/newConnection.ts
mattdean-digicatapult: we can't put this in as a parameter (we need to derive the profile using the company number from companies house) but you can look to refactor things so we avoid let

mattdean-digicatapult (approved):

PR #39, opened 2024-05-29
VR-99: converting veritable-ui to esm.

n3op2:
# Pull Request

Ideally merge after -> https://github.com/digicatapult/veritable-ui/pull/37

## Checklist

## PR Type

- [x] Bug Fix
- [x] Feature

## Linked tickets

https://digicatapult.atlassian.net/browse/VR-99

## High level description

Upgrading veritable-ui (this) to be an esm. Currently we are building using older `commonjs`. More details on this -> https://digicatapult.atlassian.net/browse/VR-99

## Operational impact

Low level - currently this services (veritable-ui) not being deployed anywhere.


src/models/__tests__/helpers/mockCompanyHouse.ts
```diff
@@ -1,14 +1,14 @@
...
+import { Env } from '../../../env.js'
 import {
   invalidCompanyNumber,
   noCompanyNumber,
   successResponse,
   validCompanyNumber,
-} from '../fixtures/companyHouseFixtures'
+} from '../fixtures/companyHouseFixtures.js'
 
-const env = container.resolve(Env)
```
mattdean-digicatapult, line 11: definitely shouldn't be using `any` here. If you need a type annotation you should be able to use `Env`

src/controllers/__tests__/helpers.ts
```diff
@@ -4,7 +4,7 @@ import pino from 'pino'
 
 import { Env } from '../../env.js'
 
-export const mockLogger = pino({ level: 'silent' })
+export const mockLogger = pino.default({ level: 'silent' })
```
mattdean-digicatapult, line 7: the `default` here is pretty ugly. You can probably change the import to 

```ts
import { default as pino } from 'pino`
```

mattdean-digicatapult (changes requested):
Looking good. Couple of small changes needed

src/controllers/__tests__/helpers.ts
n3op2: So it works for pino, but for pino-http did not work and we use it only in `server.ts` others should have been updated.

src/models/__tests__/helpers/mockCompanyHouse.ts
n3op2: ooops:) agreed and it has been updated to be itself.

mattdean-digicatapult (approved):

PR #38, opened 2024-05-29
Feature/send invitation

mattdean-digicatapult:
# Pull Request

## Checklist
- [x] Have you read Digital Catapult's [Code of Conduct](https://github.com/digicatapult/.github/blob/main/CODE_OF_CONDUCT.md)?
- [x] I have performed a self-review of my own code.
- [x] I have commented my code, particularly in hard-to-understand areas.
- [ ] I have made corresponding changes to the documentation.
- [ ] My changes generate no new warnings.
- [x] I have added tests that prove my fix is effective or that my feature works.
- [x] New and existing unit tests pass locally with my changes.

## PR Type

Please delete options that are irrelevant.

- [ ] Bug Fix
- [x] Feature
- [ ] Documentation Update
- [ ] Code style update (formatting, local variables)
- [ ] Breaking Change (fix or feature that would cause existing functionality to change)

## Linked tickets

[VR-59](https://digicatapult.atlassian.net/browse/VR-59)

## High level description

Adds connection establishment invite emails

## Detailed description

This is quite a significant PR that adds a lot of new things including:
* new database structures to accomodate invite details
* refined logic for the new connection form process including various error paths
* generation of the out of band invite in the cloudagent
* generation of the second factor PIN number and stroage in the db as an argon2 hash
* unit tests for all the new connection controllers. These were previously missed
* integration tests for creating the new connection

## Describe alternatives you've considered

PR is implemented as described in the ticket.

## Operational impact

New environment variables will need to be configured but as there's no deployment yet this doesn't have any impact

## Additional context

N/A

[VR-59]: https://digicatapult.atlassian.net/browse/VR-59?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #41, opened 2024-05-31
styling for Initiate connection request form

Ellenn-A:
# Pull Request

## Checklist

## PR Type

- [x] Bug Fix
- [x] Documentation Update
- [x] Code style update (formatting, local variables)

## Linked tickets

- https://digicatapult.atlassian.net/browse/VR-118
- https://digicatapult.atlassian.net/browse/VR-55

## High level description

Added styling and minor code changes, conserving functionality. 

## Detailed description
Added styling according to designs. 
Screenshot 2024-06-03 at 15 56 35
Screenshot 2024-06-03 at 15 56 55
Screenshot 2024-06-03 at 15 57 10


public/styles/main.css
```diff
@@ -1,19 +1,22 @@
...
   /* color pallete */
   --text-color: #45494c;
-  --text-color-secondary: #6E7079;
+  --text-color-secondary: #6e7079;
   --text-color-sub: #c9cace;
   --bg-color: #f4f5fb;
-  --secondary-color: #FFCC91;
+  --secondary-color: #ffcc91;
   --accent-color: #5670f1;
+  --light-green: #32936f;
```
mattdean-digicatapult, line 19: is this used?

public/styles/main.css
```diff
@@ -146,12 +158,12 @@ a.connections-table.icon {
   height: 14px;
   background-repeat: no-repeat;
   background-position: center;
-  background-color: transparent; 
-  background-image: url("/public/images/plus.svg");  
+  background-color: transparent;
+  background-image: url('/public/images/plus.svg');
 }
-
+/* unused??? */
```
mattdean-digicatapult, line 164: if it's unused remove

src/views/newConnection.tsx
```diff
@@ -26,7 +26,9 @@ export default class newConnectionTemplates {
         <div class="connections header">
           <span>Invite New Connection</span>
         </div>
-        <this.companyFormInput targetBox={targetBox} formStage={formStage} />
+        <div class="connections list ">
```
mattdean-digicatapult, line 29: this looks a bit silly. I assume this wasn't mean to be two classes but instead one called `connection-list`?

src/views/newConnection.tsx
```diff
@@ -43,64 +45,92 @@ export default class newConnectionTemplates {
...
-              <p>
-                {Html.escapeHtml(
-                  `Company House Number: ${params.targetBox.status === 'success' && params.targetBox.company.company_number}`
-                )}
-              </p>
-              <p>{Html.escapeHtml(`Email Address: ${params.targetBox.status === 'success' && params.email}`)}</p>
+          <this.stepper formStage={params.formStage} />
+          <div class="align-horizontally">
+            <div>
+              <div class="align-vertically ">
```
mattdean-digicatapult, line 51: This is quite the div soup. I migth need you to walk me through why we need this.

public/styles/main.css
```diff
@@ -91,6 +95,11 @@ a.connections-table.icon {
   font-size: 2rem;
   padding: 0.5rem;
 }
+.header.small {
```
mattdean-digicatapult, line 98: I'm not sure I like how we're re-suing `small` here for heading and test. If there's some shared meaning for small (say `font-size: 0.8em`) then it should be defined in it's own small class. If small means something completely different in the two cases we should probably just have `heading-small` as a single class. I get you're just following patterns elsewhere, but this is the time these patterns become fixed so I'm being a bit strict

src/views/newConnection.tsx
```diff
@@ -150,11 +185,32 @@ export default class newConnectionTemplates {
 
   public stepper = (params: { formStage: FormStage }): JSX.Element => {
     if (params.formStage === 'form') {
-      return <div>-----Step 1</div>
+      return (
+        <>
```
mattdean-digicatapult, line 189: you don't need the fragments here as there is only one child (same with the others in this function), remove

seeds/connections-seed.cjs
mattdean-digicatapult, line 1: you can revert this change I fixed it in my PR #38

mattdean-digicatapult (changes requested):

src/views/newConnection.tsx
Ellenn-A: this was sth Paul was using I left it be and created a new class `card-body` which I'll be using

public/styles/main.css
```diff
@@ -251,3 +281,92 @@ a.connections-table.icon {
...
+}
+.bar-1-3::before {
+  width: 33.33%; /* 1/3 of the bar */
+}
+.bar-2-3::before {
+  width: 66.66%; /* 2/3 of the bar */
+}
+.bar-3-3::before {
+  width: 100%; /* 3/3 of the bar */
+}
```
mattdean-digicatapult, line 367: one really small thing, it's up to you if you change. Can you change these to instead of changing width making them change the `transform` property with https://developer.mozilla.org/en-US/docs/Web/CSS/transform-function/scaleX? This will make it smoother when we animate this (which I can definitely see us doing). See also https://developer.mozilla.org/en-US/docs/Web/CSS/transform-origin

mattdean-digicatapult (approved):
I think there's some more work to be done on styling but this is a massive improvement. Approving so we can keep moving

PR #64, opened 2024-06-19
Feature/cloudagent connection sync

mattdean-digicatapult:
# Pull Request

## Checklist

## PR Type

- [ ] Bug Fix
- [x] Feature
- [ ] Documentation Update

## Linked tickets

[VR-132](https://digicatapult.atlassian.net/browse/VR-132)

## High level description

Adds connection state reconcillation with the cloudagent using the agent's websocket API

## Detailed description

This was more complex than anticipated. Let me break down the changes:

* implement an event-handler pattern that works with async event handlers `IndexedAsyncEventEmitter`
* implement the cloudagent websocket connection
* handle cloudagent connection events updating the db appropriately
* add additional cloudagent rest methods for managing connections
* minor fixes in other places so things actually work

Most of these are fairly easy to understand. The hard part to implement was the `IndexedAsyncEventEmitter`. Essentially we need a way of handling async events with a few additional properties:

1. we need to retry as missing events could leave the UI out of date indefinitely
2. we need to synchronise state on first connection. Note this synchronisation must happen asynchronously w.r.t. and further events. There therefore needs to be a cancellation mechanism to make sure that if a later event comes in on the same index that is the one that takes prescedence

Now node has the ability to catch promise rejections in async handlers as a good start but this method doesn't easily allow us to get at the event data. There's therefore a bit of type gymnastics to make it all work. 

https://github.com/digicatapult/veritable-ui/assets/29942957/046e3ebd-1acc-4dc1-81ad-7df872babc9f

Considered doing polling and not using websockets. This would have been much easier and is what I should have done. That said this is here now.

## Operational impact

Need to add new env `CLOUDAGENT_ADMIN_WS_ORIGIN` to helm chart before this is deployed.

## Additional context

Finishes off the connection establishment workflow

[VR-132]: https://digicatapult.atlassian.net/browse/VR-132?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/services/connectionEvents.ts
mattdean-digicatapult, line 1: This file is missing unit tests but I'm dubious to their value. Basically we'll just be testing the database calls happen

src/services/veritableCloudagentEvents.ts
mattdean-digicatapult, line 1: I've not unit tested this again because the tests will involve some quite intricate mocks that make the value dubious

n3op2 (commented):
almost there would like to spin up locally for final validation

src/services/veritableCloudagentEvents.ts
```diff
@@ -0,0 +1,139 @@
...
+      const connectionSeen = new Set<string>()
+
+      socket.addEventListener('message', (ev: WebSocket.MessageEvent) => {
+        this.logger.debug('WebSocket Message Received')
+        let data: WebSocketEvent
+        try {
+          data = eventParser.parse(JSON.parse(ev.data.toString()))
+        } catch (err) {
+          this.logger.warn('Unusable event received %o', ev.data)
+          return
```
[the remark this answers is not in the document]
mattdean-digicatapult: I always feel like returning like this is the correct move, especially with typescript. If I return a value like you suggest that means that the whole function now returns whatever the semantic return type of that value is. But that's not the case. Here a return just means this function returns nothing.

I do get that in older javascript we had the risk that a return without value and without semicolons might be misinterpretted but I don't think this is a problem now we have typescript and prettier

n3op2 (approved):
Sorry for taking some time, wanted to setup docker cli e.g. `docker login`.

LGTM - verified locally by spining docker compose and firing some events of the cloudagent

PR #83, opened 2024-07-02
queries list page

Ellenn-A:
# Pull Request

## Checklist

## PR Type

- [ ] Bug Fix
- [x] Feature
- [ ] Documentation Update

## Linked tickets

- https://digicatapult.atlassian.net/browse/VR-138

## High level description

Be able to view details of queries of ongoing and recent queries between an actor and  OEM or Parts Supplier , to manage  interactions and queries.

## Detailed description

- set up db table for queries with a foreign key to `connection` table `id` column (at the moment behaviour is - we can only add queries for existing connections and can only delete connections if queries are deleted - do we want to delete all associated queries if a connection is removed?) 
- set up a queries table under `Query Management` 
- for now showing full time/date...etc. do we only want date shown? 
- now the path is `Homepage > Queries > Query Management > click Query Request button > choose Query template
- filtering was conserved on the list page 
- added db seed 

Screenshot 2024-07-02 at 11 29 13

Screenshot 2024-07-02 at 11 29 33

N/A

## Operational impact

Will be able to view queries Sent and Received to other actors in a list and be able to search in them by `Company Name`

## Additional context

Reusing styling from `Connections` page a lot - I have decided not to rename, just reuse as the rename etc. should be taken care of in https://digicatapult.atlassian.net/browse/VR-169.


src/models/db/migrations/20240628090906_query_table.ts
```diff
@@ -0,0 +1,39 @@
+import { Knex } from 'knex'
+
+export async function up(knex: Knex): Promise<void> {
+  const now = () => knex.fn.now()
+
+  await knex.schema.createTable('queries', (def) => {
```
mattdean-digicatapult, line 6: We currently use singulars as table names for example `connection` so this should be `query`

src/models/db/migrations/20240628090906_query_table.ts
```diff
@@ -0,0 +1,39 @@
...
+    def.uuid('id').defaultTo(knex.raw('uuid_generate_v4()')).primary()
+    def.uuid('connection_id').notNullable()
+    def.string('company_name').notNullable()
```
mattdean-digicatapult, line 9: This is redundent as it's stored in the `connection` table. We should rely on the fk and either do a join or do two queries. Drop the column

src/models/db/migrations/20240628090906_query_table.ts
```diff
@@ -0,0 +1,39 @@
...
+        useNative: true,
+        enumName: 'query_direction',
+      })
+      .notNullable()
+    def
+      .enum('status', ['Resolved', 'Pending Your Input', 'Pending Their Input'], {
+        useNative: true,
+        enumName: 'query_status',
+      })
+      .notNullable()
```
mattdean-digicatapult, line 22: I'd prefer enum values to be lower snake case and then do a transform in the typescript.

src/models/db/migrations/20240628090906_query_table.ts
```diff
@@ -0,0 +1,39 @@
...
+    def
+      .enum('action_items', ['View Details', 'Respond'], {
+        useNative: true,
+        enumName: 'query_action_items',
+      })
+      .notNullable()
+    def.datetime('created_at').notNullable().defaultTo(now())
+    def.datetime('updated_at').notNullable().defaultTo(now())
+
+    def.foreign('connection_id').references('id').inTable('connection')
```
mattdean-digicatapult, line 32: Foregin key should cascade delete and update operations

mattdean-digicatapult (changes requested):
Just comments on db structure for now

src/views/queries.tsx
```diff
@@ -10,10 +10,13 @@ export default class QueriesTemplates {
       <Page
         title="Veritable - Queries"
         heading="Queries"
-        headerLinks={[{ name: 'Queries', url: '/queries' }]}
+        headerLinks={[
+          { name: 'Query Management', url: '/query-management' },
+          { name: 'Queries', url: '/query-management/queries' },
+        ]}
         stylesheets={['new-invite.css']}
```
mattdean-digicatapult, line 17: We shouldb't have the `new-invite` stylesheet in this page. I should have spotted this in the previous PR sorry

src/views/queriesList.tsx
```diff
@@ -0,0 +1,124 @@
...
+
+  public listPage = (queries: Query[], search: string = '') => {
+    return (
+      <Page
+        title="Veritable - Queries"
+        heading="Query Management"
+        headerLinks={[{ name: 'Query Management', url: '/query-management' }]}
+      >
+        <div
+          class="main connections"
```
mattdean-digicatapult, line 39: There's a bunch of `connection` related stuff in this file that I assume are hangovers from the original table. You have two choices:

1. generify the existing classes/ids so they are not connection specific. This may involve moving things to shared stylesheets. It may also involve extracting common utility classes
2. rename elemeents and classes in this file to be `query` rather than `connection` related. This will then involve creating new CSS which may be a lot of copying but allows us to maintain them separately

The ideal solution is likely a mixture of the two approaches

mattdean-digicatapult (changes requested):

src/views/queriesList.tsx
Ellenn-A: yh I'll have a look .. I thought this could be done in the `fixing-css` ticket, but I'll do it here

src/views/queriesList.tsx
```diff
@@ -0,0 +1,130 @@
...
+                  <tr>
+                    <td>No Queries for that search. Try again or add a new query</td>
+                  </tr>
+                ) : (
+                  queries.map((query) => (
+                    <tr>
+                      <td>{Html.escapeHtml(query.company_name)}</td>
+                      <td>{Html.escapeHtml(query.query_type)}</td>
+                      <td>{this.direction(query.status)}</td>
+                      <td>{Html.escapeHtml(query.updated_at)}</td>
```
mattdean-digicatapult, line 107: I think you're date issue is here. Basically this will render a time based on the locale of the server which we don't want. I think the best fix is to just use the [time](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/time) element and set the `datetime` attribute to the UTC time here. You may have to do some timezon schenanigans to make sure it displays correctly in the browser

src/controllers/queries/index.ts
```diff
@@ -8,6 +8,33 @@ import QueriesTemplates from '../../views/queries.js'
...
+type QuerySubset = {
+  updated_at: Date
+  connection_id: string
+  status: 'resolved' | 'pending_your_input' | 'pending_their_input'
+  id: string
+  created_at: Date
+  query_type: string
+}
+
+type Connection = {
```
mattdean-digicatapult, line 27: Does this type not exist somewhere else? Can you just import it?

Ellenn-A: not that I could see ..the closest there is is the connection interface in views>connection.tsx but that one is missing id `interface connection {
  company_name: string
  status: ConnectionStatus
}`

mattdean-digicatapult: maybe
```ts
import type { ConnectionRow } from '../../models/db/types.js`
```
Likely you can export the Query type there as well maybe so it's located in a more shared place assoicated with where it comes from?

Ellenn-A: thank you! I'll change it

src/models/db/migrations/20240628090906_query_table.ts
```diff
@@ -0,0 +1,27 @@
...
+    def.uuid('id').defaultTo(knex.raw('uuid_generate_v4()')).primary()
+    def.uuid('connection_id').notNullable()
+    def.string('query_type').notNullable()
+    def
+      .enum('status', ['resolved', 'pending_your_input', 'pending_their_input'], {
+        useNative: true,
+        enumName: 'query_status',
+      })
+      .notNullable()
+      .notNullable()
```
mattdean-digicatapult, line 16: duplicate

mattdean-digicatapult (approved):

PR #86, opened 2024-07-03
Combining two forms into one, since they both for a new connection

n3op2:
# Pull Request

## Checklist

## PR Type

- [x] Code style update (formatting, local variables)

## Linked tickets

https://digicatapult.atlassian.net/jira/software/c/projects/ENG/boards/206?quickFilter=175&selectedIssue=VR-153&sprints=620

## High level description

Abstracted two form into one

## Detailed description

We used to have two different classes for each form e.g. sending invitation and verifying.

## Operational impact

None - as far as I'm aware


src/controllers/connection/__tests__/index.test.ts
```diff
@@ -16,7 +16,7 @@ describe('ConnectionController', () => {
       let { dbMock, mockLogger, templateMock } = withConnectionMocks()
       const controller = new ConnectionController(dbMock, templateMock, mockLogger)
       const result = await controller.listConnections().then(toHTMLString)
-      expect(result).to.equal('list_foo-verified_list')
+      expect(result).to.equal('list_"foo"-verified_list')
```
Ellenn-A, line 19: why is the foo in  "" now? (I just can't find where it came from)

n3op2: This is from `JSON.stringify()` did not want to overdue with logic as this is mainly to increase assertation. A good attached a good example
Screenshot 2024-07-03 at 15 10 19

Ellenn-A (commented):
looks good to me

src/views/newConnection/newInvite.tsx
mattdean-digicatapult, line 1: I don't see the value in this change sorry. There's little to no overlap between the two forms apart from what is already held in the base class which you've for some reason left in (even though now only one thing is derived from it). All this PR does is move code from one form to the other. 

Can I understand the motivation for amalgamating these forms into one beyond just you don't like the code structure?

mattdean-digicatapult (changes requested):

n3op2:
this is a duplicate, and also since this refactor is no longer relevant other commits have been pushed to this branch

PR #126, opened 2024-08-13
Feat/response to query carbon emission

Ellenn-A:
# Pull Request

## Checklist

## PR Type

- [ ] Bug Fix
- [x] Feature
- [ ] Documentation Update

## Linked tickets

- https://digicatapult.atlassian.net/jira/software/c/projects/ENG/boards/206?quickFilter=175&selectedIssue=VR-174
- https://digicatapult.atlassian.net/jira/software/c/projects/ENG/boards/206?quickFilter=175&selectedIssue=VR-168
- https://digicatapult.atlassian.net/jira/software/c/projects/ENG/boards/206?quickFilter=175&selectedIssue=VR-193

## High level description

Added form and methods for replying to a received query.

## Detailed description

Updated seed files.

Implemented a frontend for replying to an existing query. In this scenario we have already received a query and it is in our database - we want to respond to it. 
Note that an addition has been made in submitting a query - added a query-id which corresponds to the id of the sender which will be used by the responder when sending a response so this can be matched to a query. 

N/A


src/services/drpcEvents/__tests__/index.test.ts
```diff
@@ -484,6 +500,339 @@ describe('DrpcEvents', function () {
...
+      it('should get the relevant connection correctly', function () {
+        const stub = mocks.dbMock.get
+        expect(stub.callCount).to.equal(1)
+        expect(stub.firstCall.args).to.deep.equal([
+          'connection',
+          { agent_connection_id: 'invalid-agent-connection-id' },
+        ])
+      })
+
+      it('should not call any other db methods or submitDrpcResponse', function () {
```
n3op2, line 727: maybe assert response as well, that is basically a fixture?

n3op2 (commented):
just a few minors, overall looks like you have done a nice job!

src/models/db/migrations/20240808125209_query_tables_additional_column.ts
```diff
@@ -0,0 +1,15 @@
+import type { Knex } from 'knex'
+
+export async function up(knex: Knex): Promise<void> {
+  await knex.schema.alterTable('query', (def) => {
+    def.uuid('query_id_for_response').defaultTo(null)
```
mattdean-digicatapult, line 5: I'm thinking we should also add a column for query role so We can tell if we've asked the question or been asked. Maybe something like `role` with an enum with values `requester` and `responder`

src/controllers/queries/index.ts
```diff
@@ -184,6 +202,116 @@ export class QueriesController extends HTMLController {
     )
   }
 
+  /**
+   * Retrieves the query response page
+   */
+  @SuccessResponse(200)
+  @Get('/scope-3-carbon-consumption-response/{queryId}')
```
mattdean-digicatapult, line 209: I think we should clean up the path a bit here. Maybe something like `/scope-3-carbon-consumption/{queryId}/response`

src/controllers/queries/index.ts
```diff
@@ -184,6 +202,116 @@ export class QueriesController extends HTMLController {
...
+        query.details['productId']
+      )
+    )
+  }
+
+  /**
+   * Submits the query response page
+   */
+  @SuccessResponse(200)
+  @Post('/scope-3-carbon-consumption-response/submit')
```
mattdean-digicatapult, line 238: `/scope-3-carbon-consumption/{queryId}/response`

src/controllers/queries/index.ts
```diff
@@ -184,6 +202,116 @@ export class QueriesController extends HTMLController {
...
+  public async scope3CarbonConsumptionResponseSubmit(
+    @Body()
+    body: {
+      companyId: UUID
+      queryId: UUID
```
mattdean-digicatapult, line 243: queryId should be in the path

src/controllers/queries/index.ts
```diff
@@ -184,6 +202,116 @@ export class QueriesController extends HTMLController {
...
+    this.logger.debug('query page requested')
+    //retrieve query
+    const queries = await this.db.get('query', { id: queryId })
+    if (queries.length < 1) {
+      throw new Error(`There has been an issue retrieving the query.`)
+    }
+    const query = queries[0]
+    const connections = await this.db.get('connection', { id: query.connection_id })
+    if (connections.length < 1) {
+      throw new Error(`There has been an issue retrieving the connection.`)
```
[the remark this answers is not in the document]
mattdean-digicatapult: nope this is a `InvalidInputError`. `404` is explicitly for path resources

src/services/drpcEvents/index.ts
```diff
@@ -50,8 +52,11 @@ export default class DrpcEvents {
   }
 
   private async handleRequestReceived(request: DrpcRequest, agentConnectionId: string) {
-    if (request.method === 'submit_query_request') {
-      return await neverFail(this.handleSubmitQueryRequest(request, agentConnectionId))
+    switch (request.method) {
+      case 'submit_query_request':
+        return await neverFail(this.handleSubmitQueryRequest(request, agentConnectionId))
+      case 'submit_query_response':
+        return await neverFail(this.handleSubmitQueryResponse(request, agentConnectionId)) //new method I am making
```
mattdean-digicatapult, line 59: remove comment

src/services/drpcEvents/index.ts
```diff
@@ -146,4 +153,85 @@ export default class DrpcEvents {
...
+      this.logger.info(
+        'DRPC response (%s) received on connection %s of method %s',
+        request.id,
+        agentConnectionId,
+        request.method
+      )
+
+      let params: SubmitQueryRPCParams
+      try {
+        params = submitQueryRpcParams.parse(request.params)
```
mattdean-digicatapult, line 169: you shouldn't reuse the same parser here. You should only need the `queryId` and the `emissions` I think.

src/services/drpcEvents/index.ts
mattdean-digicatapult: I also need to think if `neverFail` is appropriate here

src/services/drpcEvents/index.ts
```diff
@@ -146,4 +153,85 @@ export default class DrpcEvents {
...
+      if (!connection) {
+        this.logger.warn('Invalid queryId for drpc message %s', params.queryIdForResponse)
+        return
+      }
+      //update corresponding query
+      this.logger.warn(queryRow)
+      const [query] = await this.db.update(
+        'query',
+        { id: queryRow.id },
+        { query_response: params.emissions, status: 'resolved' }
```
mattdean-digicatapult, line 197: I think we can simplify this to `{ response: params.emissions, status: 'resolved' }`

src/services/drpcEvents/index.ts
```diff
@@ -146,4 +153,85 @@ export default class DrpcEvents {
...
+
+      queryId = query.id
+      const result = {
+        state: 'accepted',
+      }
+      await this.cloudagent.submitDrpcResponse(request.id, { result })
+      await this.db.insert('query_rpc', {
+        query_id: query.id,
+        method: 'submit_query_response',
+        role: 'server', //am I client here?
```
mattdean-digicatapult, line 208: `server` is correct

src/models/db/migrations/20240808125209_query_tables_additional_column.ts
Ellenn-A: I thought we can know this based on whether we have received the 'query-id-for-response' == a query_id we arebeing asked to respond to

src/services/drpcEvents/index.ts
Ellenn-A: Yeah I wasn't sure if we'd want to also keep the original query just in case there is a mix up or sth so we can prove that whoever is answering is providing an answer only to that one query...but if you think it's unnecessary then I'm happy to remove

src/services/drpcEvents/index.ts
Ellenn-A: but we need to make sure that the status has been updated? Unless we want to remove the status field?

src/models/db/migrations/20240808125209_query_tables_additional_column.ts
mattdean-digicatapult: We probably can, but this is getting to the point where it's difficult to tell and I feel these sorts of inferences require explicit knowledge of how the system works. A column for this would be useful and simpler. We can also filter on it in the UI then and in the event handler much more simply

src/services/drpcEvents/index.ts
mattdean-digicatapult: yep please change to using a dedicated parser

src/services/drpcEvents/index.ts
mattdean-digicatapult: yep misread. Leave as is

src/models/db/migrations/20240808125209_query_tables_additional_column.ts
Ellenn-A: I shall add the column

seeds/connections-queries-seed.js
```diff
@@ -33,6 +33,7 @@ export async function prepareVariants() {
         status: 'verified_both',
         company_number: i.toString().padStart(8, '0'),
         company_name: `VER1553 company_name Name ${i.toString()}`,
+        agent_connection_id: `11dc8dac-1988-49f2-8761-ffedcbd81a91`,
```
n3op2, line 36: don't we want to give a unique `agent_connection_id` for each record?

src/controllers/queries/index.ts
```diff
@@ -184,6 +203,114 @@ export class QueriesController extends HTMLController {
...
+      if (!maybeResponse) {
+        return await this.handleError(queryRow, connection)
+      }
+      rpcResponse = maybeResponse
+    } catch (err) {
+      return await this.handleError(queryRow, connection, undefined, err)
+    }
+    const { result, error, id: rpcId } = rpcResponse
+
+    await this.db.insert('query_rpc', {
```
n3op2, line 289: just wonder why it's not all under one `try/catch` block?

src/services/drpcEvents/__tests__/index.test.ts
```diff
@@ -484,6 +502,339 @@ describe('DrpcEvents', function () {
...
+      it('should not call any db or cloudagent methods', function () {
+        const stubs = [
+          mocks.cloudagentMock.submitDrpcResponse,
+          mocks.dbMock.get,
+          mocks.dbMock.update,
+          mocks.dbMock.insert,
+        ]
+        for (const stub of stubs) {
+          expect(stub.callCount).to.equal(0)
+        }
```
n3op2, line 606: like this implementation

n3op2 (commented):
Screenshot 2024-08-21 at 09 48 30
pretty sure I've updated in my PR, maybe is just a matter of rebase

src/controllers/queries/index.ts
```diff
@@ -184,6 +203,114 @@ export class QueriesController extends HTMLController {
...
+    if (!connection) {
+      throw new NotFoundError(`Invalid connection ${body.companyId}`)
+    }
+    if (!connection.agent_connection_id || connection.status !== 'verified_both') {
+      throw new NotFoundError(`Cannot query unverified connection`)
+    }
+    const [queryRow] = await this.db.get('query', { id: queryId })
+    if (!queryRow.response_id) {
+      throw new NotFoundError(`Missing queryId to respond to.`)
+    }
```
mattdean-digicatapult, line 264: These should not be `NotFoundError`. `NotFoundError` is a `404` response which should be reserved for non existant path resources. In this case that's only for `queryId` values not found in the database. This also happens to be a case we're not catching.

In short we need 

* an extra check for the truthyness of `queryRow`. Is falsey this should throw `NotFoundError`.
* replace the existing throws with a `InvalidInputError`

src/controllers/queries/index.ts
mattdean-digicatapult: I htink this is sensible. We expect the `submitDrpcRequest` call might fail and want to handle that specifically. Errors elsewhere would be unexpected and can be handled by the global handler

src/services/drpcEvents/index.ts
```diff
@@ -146,4 +161,90 @@ export default class DrpcEvents {
...
+        return
+      }
+
+      //update corresponding query
+      this.logger.warn(queryRow)
+      const [query] = await this.db.update(
+        'query',
+        { id: queryRow.id },
+        { query_response: params.emissions, status: 'resolved' }
+      ) //do we want to save the full params object?
```
mattdean-digicatapult, line 211: remove comment

src/services/drpcEvents/index.ts
```diff
@@ -146,4 +161,90 @@ export default class DrpcEvents {
...
+
+      if (queryId !== null) {
+        await this.db.update(
+          'query',
+          { id: queryId },
+          {
+            status: 'errored',
+          }
+        )
+      }
```
mattdean-digicatapult, line 241: I've thought about this and we should remove this update. We don't want to invalidate a query because of an unknown error. I think the lesser evil here is to leave the ability for the other side to provide an answer

mattdean-digicatapult (changes requested):
Marking these as changes required. @n3op2 can you prioritise these changes and updating the tests. Shouldn't take long.

If I'm still away when this finishes feel free to dismiss this @dblane-digicatapult in order to get this merged once the devs agree these are done

src/controllers/queries/index.ts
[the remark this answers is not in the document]
n3op2: just double checked, that has been updated,

src/models/db/migrations/20240808125209_query_tables_additional_column.ts
n3op2: do you think we should add a defaultTo('requester') value? or make it nullable? 

```ts
// currently
    def
      .enum('role', ['requester', 'responder'], {
        useNative: true,
        enumName: 'query_role',
      })
      .notNullable()
```

src/services/drpcEvents/index.ts
n3op2: It looks like this has been updated 

```js
      let params: SubmitQueryResponseRPCParams
      try {
        params = submitQueryResponseRpcParams.parse(request.params)
```

src/controllers/queries/index.ts
n3op2: yea, makes sense, otherwise would say we do have a middleware that should handle all exceptions and for unknown - return 5xx

src/services/drpcEvents/index.ts
n3op2: Agreed, removed and left a comment in the code in regards to fire and forget.

src/services/drpcEvents/index.ts
n3op2: removed

src/services/drpcEvents/index.ts
```diff
@@ -146,4 +161,90 @@ export default class DrpcEvents {
...
+        return
+      }
+      const [connection] = await this.db.get('connection', { agent_connection_id: agentConnectionId })
+      if (!connection) {
+        this.logger.warn('Invalid connection for drpc message %s', agentConnectionId)
+        return
+      }
+      //find corresponding query based on queryIdForResponse provided by responder,
+      //this should match a query id in our db
+      const [queryRow] = await this.db.get('query', { id: params.queryIdForResponse })
```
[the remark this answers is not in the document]
n3op2: also updated database to have `defaultTo('requester`)`

src/controllers/queries/index.ts
n3op2: Should be updated, not sure if you will like it, but broken down into 
- queryRow = NotFound
- invalid inputs for others like "status", "response_id" and etc.

mattdean-digicatapult (approved):

PR #178, opened 2024-09-20
Feat/e2e connections and queries

Ellenn-A:
# Pull Request

## Checklist

## PR Type

- [ ] Bug Fix
- [x] Feature
- [x] Documentation Update

## Linked tickets

https://digicatapult.atlassian.net/jira/software/c/projects/ENG/boards/206?quickFilter=175&selectedIssue=VR-205&sprints=814

## High level description

This is the first of a series of tests on this ticket. This one establishes connection between Alice and Bob. 

## Detailed description

Establishes connection between Alice and Bob, originating from Alice. This is done via ui. The flow of the test can be followed in file `connection-alice-to-bob.spec.ts`. The email transport in these tests is set to SMTP_EMAIL (meaning a test server `smtp4dev` is being used instead of simply printing the pin and invitation. 

The structure is of one test with multiple `test.step` separating different steps. This was done so that our flow runs sequentially. Once a step or a check in a test flow fails, the tests do not continue running. 

N/A

## Operational impact

N/A

## Additional context

N/A


test/e2e/helpers/utils.ts
```diff
@@ -0,0 +1,3 @@
+export function sleep(ms: number): Promise<void> {
+  return new Promise((resolve) => setTimeout(resolve, ms))
+}
```
n3op2, line 3: maybe add this along other test utils? since it has it's own dedicater forlder "helpers" already

```
test/
├── e2e
│   └── homepage-redirect.spec.ts
├── helpers
│   ├── cloudagent.ts
│   ├── companyHouse.ts
│   ├── connection.ts
│   ├── db.ts
│   ├── fixtures.ts
│   ├── logger.ts
│   ├── routeHelper.ts
│   ├── smtp.ts
│   └── util.ts <<<< here maybe?
...
```

test/e2e/helpers/smtpEmails.ts
```diff
@@ -0,0 +1,182 @@
+import { expect } from '@playwright/test'
```
n3op2, line 1: same for this helper, would be nice to keep all in one place e.g. `tests/helpers/` folder

test/e2e/connection-alice-to-bob.spec.ts
```diff
@@ -0,0 +1,218 @@
...
+    await test.step('Alice invites Bob to connect', async () => {
+      await page.goto('http://localhost:3000')
+      const inviteUrl = page.url()
+      if (
+        inviteUrl.includes(
+          'http://localhost:3080/realms/veritable/protocol/openid-connect/auth?response_type=code&client_id=veritable-ui&redirect_uri=http'
+        )
+      ) {
+        await page.fill('#username', 'name')
+        await page.fill('#password', 'password')
```
n3op2, line 58: maybe this is something that could go to beforeEach?

test/e2e/connection-alice-to-bob.spec.ts
```diff
@@ -0,0 +1,218 @@
...
+      await page.goto('http://localhost:3001')
+      const bobUrl = page.url()
+      // Login if redirected
+      if (
+        bobUrl.includes(
+          'http://localhost:3080/realms/veritable/protocol/openid-connect/auth?response_type=code&client_id=veritable-ui&redirect_uri=http'
+        )
+      ) {
+        await page.fill('#username', 'name')
+        await page.fill('#password', 'password')
```
n3op2, line 125: same here

test/e2e/connection-alice-to-bob.spec.ts
```diff
@@ -0,0 +1,218 @@
...
+      await page.goto('http://localhost:3000')
+      const aliceUrl = page.url()
+      // Login if redirected
+      if (
+        aliceUrl.includes(
+          'http://localhost:3080/realms/veritable/protocol/openid-connect/auth?response_type=code&client_id=veritable-ui&redirect_uri=http'
+        )
+      ) {
+        await page.fill('#username', 'name')
+        await page.fill('#password', 'password')
```
n3op2, line 187: same here

test/e2e/connection-alice-to-bob.spec.ts
```diff
@@ -0,0 +1,218 @@
...
+      await page.waitForURL('**/realms/veritable/login-actions/registration**')
+
+      await page.fill('#username', 'name')
+      await page.fill('#password', 'password')
+      await page.fill('#password-confirm', 'password')
+      await page.fill('#email', 'email@testmail.com')
+      await page.fill('#firstName', 'name')
+      await page.fill('#lastName', 'lastname')
+      await page.click('input[type="submit"][value="Register"]')
+      await page.waitForURL('http://localhost:3000')
```
n3op2, line 46: maybe register could be abstracted as well since this is something we will need for each suite? to be fair not a major so this does not block me from approving

test/e2e/connection-alice-to-bob.spec.ts
```diff
@@ -0,0 +1,218 @@
...
+
+      await page.fill('#new-connection-invite-input-pin', pinForBob)
+      await page.click('button[type="submit"][name="action"][value="submitPinCode"]')
+
+      await page.waitForURL('**/connection')
+    })
+
+    await test.step('Retrieve pin for Alice', async () => {
+      const newAdminEmail = await findNewAdminEmail(adminEmailId)
+      pinForAlice = await extractPin(newAdminEmail.id)
```
n3op2, line 172: why it can't be like?

```ts
const pinForAlice = await.extractPin()
```

test/e2e/connection-alice-to-bob.spec.ts
```diff
@@ -0,0 +1,218 @@
+import { expect, test } from '@playwright/test'
+import { checkEmails, extractInvite, extractPin, findNewAdminEmail } from './helpers/smtpEmails'
+
+test.describe('Connection from Alice to Bob', () => {
+  let context: any
+  let page: any
+  let adminEmailId: string
+  let invite: string | null
+  let pinForBob: string | null
+  let pinForAlice: string | null
```
n3op2, line 10: quite a few lets, wonder if they are all needed

n3op2 (commented):
Leaving some comments, mainly around helpers also noticed some new methods e.g. we use test object where are blocks like `describe` `it` `beforeAll` and etc so I might be missing some context and if some comments do not make sense - please ignore

test/e2e/helpers/utils.ts
Ellenn-A: I did not want to mix utils for integration and e2e tests

test/e2e/connection-alice-to-bob.spec.ts
Ellenn-A: I'm not sure I understand this question

test/e2e/connection-alice-to-bob.spec.ts
Ellenn-A: ` let context: any // shared between all pages
  let page: any 
  let adminEmailId: string // to exclude this when looking for Alice's email containing PIN
  let invite: string | null // for Bob -> we extract in one step and use in another so saving it here 
  let pinForBob: string | null // same as above 
  let pinForAlice: string | null // also extracted in one step and used in another - so saving in this var`

test/e2e/helpers/utils.ts
mattdean-digicatapult: is this function used?

test/e2e/connection-alice-to-bob.spec.ts
```diff
@@ -0,0 +1,188 @@
...
+      expect(positiveFeedback).not.toBeNull()
+      const feedback = await page.$('#new-connection-feedback')
+      expect(feedback).not.toBeNull()
+
+      const feedbackText = await feedback?.textContent()
+      expect(feedbackText).toContain('Registered Office Address')
+      expect(feedbackText).toContain('DIGITAL CATAPULT')
+
+      await page.click('button[type="submit"][name="action"][value="createConnection"]')
+      await page.waitForTimeout(3000)
```
mattdean-digicatapult, line 140: is this in essence a `sleep`? Can we replace this with something less flakey?

test/e2e/connection-alice-to-bob.spec.ts
```diff
@@ -0,0 +1,188 @@
...
+      await page.fill('#username', 'name')
+      await page.fill('#password', 'password')
+      await page.click('input[type="submit"][value="Sign In"]')
+      await page.waitForURL('http://localhost:3000')
+    }
+  })
+
+  test.afterEach(async () => {
+    await page.close()
+  })
```
mattdean-digicatapult, line 54: These look like test preconditions. Can we move these into the test harness? functions like `withRegisteredAccount` `withLoggedInUser` feel appropriate

mattdean-digicatapult (commented):
Mostly looking good. Need to move some of the preconditions of the test into a re-usable test harness. Also need to look at if any of the timeouts can be removed. I'd also like to maker sure we can rely on this in CI which I think is still an issue?

docker-compose.e2e.yml
```diff
@@ -0,0 +1,63 @@
...
+      - 3000:3000
+    command: >
+      sh -c "
+        npm i -g pino-colada
+        node ./node_modules/.bin/knex migrate:latest
+        npm start | pino-colada"
+    # env_file:
+    # - docker/test.env
+    # - .env
+    environment:
```
mattdean-digicatapult, line 26: need to set

```
- COMPANY_PROFILE_API_KEY=${VERITABLE_COMPANY_PROFILE_API_KEY}
```

docker-compose.yml
```diff
@@ -73,25 +73,25 @@ services:
         npm i -g pino-colada
         node ./node_modules/.bin/knex migrate:latest
         npm start | pino-colada"
-    env_file:
-      - docker/test.env
-      - .env
+    # env_file:
+    # - docker/test.env
+    # - .env
```
mattdean-digicatapult, line 78: need to set

```
- COMPANY_PROFILE_API_KEY=${VERITABLE_COMPANY_PROFILE_API_KEY}
```

mattdean-digicatapult (commented):
Spotted a couple of places you need to pass through the api key

docker-compose.yml
```diff
@@ -103,6 +103,9 @@ services:
       - SMTP_PASS=
       - SMTP_PORT=25
       - SMTP_USER=
+      - EMAIL_TRANSPORT=${VERITABLE_EMAIL_TRANSPORT:-STREAM}
+      - SMTP_SECURE=false
+      - COMPANY_PROFILE_API_KEY=${VERITABLE_COMPANY_PROFILE_API_KEY:-COMPANY_PROFILE_API_KEY}
```
mattdean-digicatapult, line 108: I think this will break things locally as bob won't be able to call the API unless you set `VERITABLE_COMPANY_PROFILE_API_KEY` locally? `COMPANY_PROFILE_API_KEY` isn't a valid value obviously and probably we shouldn't be setting an invalid default

docker-compose.e2e.yml
```diff
@@ -0,0 +1,64 @@
...
+      - SMTP_PASS=
+      - SMTP_PORT=25
+      - SMTP_USER=
+      - EMAIL_TRANSPORT=SMTP_EMAIL
+      - API_SWAGGER_TITLE=Alice
+      - API_SWAGGER_BG_COLOR=#ff3131
+      - COMPANY_HOUSE_API_URL=https://api.company-information.service.gov.uk
+      - DEMO_MODE=true
+      - SMTP_SECURE=false
+      - COMPANY_PROFILE_API_KEY=${VERITABLE_COMPANY_PROFILE_API_KEY:-COMPANY_PROFILE_API_KEY}
```
mattdean-digicatapult, line 56: same here, but less important as this is just for the e2e tests

docker-compose.integration.yml
mattdean-digicatapult, line 1: is there a reason for breaking this out to it's own compose file?

playwright.config.ts
```diff
@@ -33,6 +39,7 @@ export default defineConfig({
       use: { ...devices['Desktop Firefox'] },
     },
 
+    // Webkit disabled for now due to a bug
```
mattdean-digicatapult, line 42: does webkit work now?

test/e2e/connection-alice-to-bob.spec.ts
```diff
@@ -0,0 +1,184 @@
...
+  let pinForAlice: string | null
+
+  const baseUrlAlice = process.env.VERITABLE_ALICE_PUBLIC_URL || 'http://localhost:3000'
+  const baseUrlBob = process.env.VERITABLE_BOB_PUBLIC_URL || 'http://localhost:3001'
+  const smtp4devUrl = process.env.VERITABLE_SMTP_ADDRESS || 'http://localhost:5000'
+  const { host, port } = getHostPort(smtp4devUrl)
+  if (host === null || port === null) {
+    throw new Error(`Unspecified smtp4dev host or port ${smtp4devUrl}`)
+  }
+  console.log(process.env.VERITABLE_ALICE_PUBLIC_URL)
```
mattdean-digicatapult, line 20: lots of console logs to cleanup

.github/workflows/test.yml
```diff
@@ -90,72 +93,29 @@ jobs:
...
-          SMTP_HOST: localhost
-          SMTP_PASS: ''
-          SMTP_PORT: 2525
-          SMTP_USER: ''
-      - name: Install Playwright Browsers
-        run: npx playwright install --with-deps
-      - name: Run Playwright tests
-        run: npm run test:playwright
+          VERITABLE_COMPANY_PROFILE_API_KEY: ${{ secrets.COMPANIES_HOUSE_API }}
+          VERITABLE_E2E_OUT_DIR: /home/runner/work/_temp/
```
mattdean-digicatapult, line 112: you should be able to template this in as `${{ runner.temp }}`. Just in case github ever changes things

docker-compose-test.yml
mattdean-digicatapult, line 1: The filenaming here is a bit random. My temptation is to reduce this to:

1. `docker-compose.yml` - Deploys two full personas bob and charlie, a persona for Alice except for her ui (so we can run that locally), `keycloak` and `smtp4dev`. We should then move to just using `smtp` as our transport all the time.
2. `docker-compose.e2e.yml` - Deploys `docker-compose.yml` along with the `veritable-ui-alice` container and the `e2e-tests` container. For the local run I'd be tempted to set a default `scale` for the test container to `0` and then that can be overridden in ci to run the tests in the container

I think the ideal setup then splits out the shared stuff for a persona to a separate compose file but we can worry about that later

playwright.config.ts
Ellenn-A: yes, passed in all 3

docker-compose.integration.yml
Ellenn-A: so for integration tests we want the smtp4dev, but not Alice...that's why

mattdean-digicatapult: But could we not just move this back into the regular `docker-compose.yml`? Is there a circumstance we don't want the `smtp4dev` service running?

playwright.config.ts
mattdean-digicatapult: fabulous, I assume this is just part of the cleanup then

docker-compose.integration.yml
Ellenn-A: good point I guess if you're bringing up everything e.g. for local testing your transport will most likely be set to 'STREAM' so then the smtp4dev would be obsolete

test/e2e/connection-alice-to-bob.spec.ts
```diff
@@ -0,0 +1,184 @@
...
+      await page.click('button[type="submit"][name="action"][value="createConnection"]')
+      await page.waitForTimeout(3000)
+
+      // Submit pin
+      await page.waitForURL('**/pin-submission')
+      const pinUrl = page.url()
+      // const urlPattern = /http:\/\/veritable-ui-bob:3000\/connection\/[0-9a-fA-F-]{36}\/pin-submission/
+      // const urlPattern = /http:\/\/localhost:3001\/connection\/[0-9a-fA-F-]{36}\/pin-submission/
+
+      const escapedBaseUrlBob = baseUrlBob.replace(/[-/\\^$*+?.()|[\]{}]/g, '\\$&')
```
Ellenn-A, line 129: regex weirdness @mattdean-digicatapult

mattdean-digicatapult (dismissed):

mattdean-digicatapult (dismissed):

mattdean-digicatapult (approved):

mattdean-digicatapult (dismissed):

PR #186, opened 2024-09-30
Make smtp envs optional

mattdean-digicatapult:
# Pull Request

## Checklist

## PR Type

- [x] Bug Fix

## Linked tickets

VR-226

## High level description

Removes need for smtp environment variables to be specified if not using smtp transport

## Detailed description

Firstly the scary number of files is because I've broken up the `env.ts` into several modules. Thus imports have changed in lots of places. It shouldn't be too bad to review but sorry it was necessary

I've broken out the different config sections into separate Env classes that each load based on the presence of other env values. This works quite well and also is quite scalable for other optional envs.

The test setup is a bit convoluted due to how envalid handles errors. There's a bit of a dance with tsyringe to inject correctly the test envs and a custom reporter so mocha doesn't get kicked out by envalid when a test fails.

N/A

## Operational impact

None

## Additional context

None


src/env/index.ts
```diff
@@ -0,0 +1,22 @@
...
+import { singleton } from 'tsyringe'
+
+import { DEFAULT_CONFIG, DEFAULT_KEYS, defaultConfig, loadEnvAndOptions, PartialEnv } from './common.js'
+
+export type { PartialEnv } from './common.js'
+export { InitEnv } from './init.js'
+export { SmtpEnv } from './smtp.js'
+
+@singleton()
+export class Env<KS extends DEFAULT_KEYS = DEFAULT_KEYS> implements PartialEnv<KS> {
```
Ellenn-A, line 11: what is the relation of InitEnv to Env?

mattdean-digicatapult: `InitEnv` is used by the init script not by the main application https://github.com/digicatapult/veritable-ui/blob/d089e533b3725df5261042223892fc50e6726f9f/src/init.ts

Ellenn-A: yes I saw, so that means the initEnv is just exclusively for that and the actual Env is for runtime envars/ stuff after the init has happened?

Ellenn-A (approved):

PR #188, opened 2024-10-02
A new page - Credentials list

n3op2:
# Pull Request

## Checklist

## PR Type

- [x] Feature

## Linked tickets

https://digicatapult.atlassian.net/browse/VR-183

## High level description

Adding a new page to the veritable-ui portal. This page mainly is just a table that renders a list of connection.

## Detailed description
 
Screenshot 2024-10-02 at 13 21 37
Screenshot 2024-10-02 at 13 21 30
Screenshot 2024-10-02 at 13 21 02
Screenshot 2024-10-02 at 13 18 49

## Operational impact

low or none, a new page/feature does not interact with other parts just cloiud agent


public/styles/shared.css
```diff
@@ -7,8 +7,9 @@
 
   &:hover,
   &:focus-visible {
-    filter: blur(1px);
-    opacity: 1;
+    filter: grayscale(100%) !important;
+    opacity: 0.3 !important;
+    filter:blur(2) p !important;
```
mattdean-digicatapult, line 12: This must be wrong. First there must be a better way than `!important` to do this. Second you're defining `filter` twice so only the second one will apply anyway

src/views/common.tsx
```diff
@@ -167,3 +179,75 @@ export const Page = (props: PropsWithChildren<PageProps>): JSX.Element => (
     </html>
   </>
 )
+
+// this method was used in three different files, if no objections would like to leave along other shared items
```
mattdean-digicatapult, line 183: Yep sounds sensible. Just remove the comment please

src/controllers/credentials/index.ts
```diff
@@ -0,0 +1,51 @@
...
+            updated_at: new Date(),
+            company_name: cred?.credentialAttributes[0].name === 'company_name' && cred.credentialAttributes[0].value,
+            company_number:
+              cred.credentialAttributes[1].name === 'company_number' && cred.credentialAttributes[1].value,
+          }
+        }
+      })
+      .filter((cred) =>
+        search !== '' ? cred?.company_name.toString().toLowerCase().includes(search.toLowerCase()) : true
+      )
```
mattdean-digicatapult, line 41: Couple of comments here. This assumes that we're looking at only this type of credential and doesn't check that this alligns with a connection in the ui database. Can I suggest two improvements:

1. Filter this list on credentials that are of type `COMPANY_DETAILS`. That way we're not misinterpretting other credentials accidentally if they happen to exist
2. Filter this list based on the other sides connection id. You can probably fetch all connections first and build a `Set` of the `agent_connection_id` values. Then check for membership of the set in a filter

Also a question, do we want to only filter credentials in certain states. That way we don't show things in flight? Do we know how to display each credential status?

src/views/credentials/index.tsx
```diff
@@ -0,0 +1,136 @@
...
+                <th>
+                  <span>{name || 'unknown'}</span>
+                  <a class="list-filter-icon icon disabled" />
+                </th>
+              ))}
+            </thead>
+            <tbody id="search-results">
+              {credentials.length == 0 ? (
+                <tr>
+                  <td>No Credentials for that search. Try again or add a new credential</td>
```
mattdean-digicatapult, line 101: Can we simplify the copy here to `No credentials found`

src/views/credentials/index.tsx
```diff
@@ -0,0 +1,136 @@
...
+                </tr>
+              ) : (
+                credentials.map((cred) => {
+                  if (!cred || !cred.credentialAttributes) throw new Error('TODO update error')
+                  return (
+                    <tr>
+                      <td>{Html.escapeHtml(cred.credentialAttributes[0].value)}</td>
+                      <td>{Html.escapeHtml('Supplier credentials')}</td>
+                      <td>{this.roleToDirection(cred.role)}</td>
+                      <td>{statusToClass(cred.state as ConnectionStatus)}</td>
```
mattdean-digicatapult, line 111: This looks incorrect. Your passing a credential status as a connection status and re-using the same transform. Does that really work?

mattdean-digicatapult (changes requested):
Looking very good. Mostly just a change to limit which credentials we're displaying whilst we're not doing this quite properly

src/views/credentials/index.tsx
n3op2: I dropped all states into one, mmm.. it does but would need to confirm across the board. since 3 files have had their own (status/State).do you think I should revert?

public/styles/shared.css
n3op2: 🤦 indeed, should have been  `filter:blur(2), grayscale(100%) !important;` mmm.... might have some unstagged stuff hanging

src/views/credentials/index.tsx
mattdean-digicatapult: There's at least two places where `statusToClass` was duplicated so I don't mind you doing that. I'm more worried that this particular one seems wrong

src/controllers/credentials/index.ts
n3op2: Good points and agent connection id is basically our source of truth
> Also a question, do we want to only filter credentials in certain states. That way we don't show things in flight? Do we know how to display each credential status?

indeed we do have quite a states of cred until we reach done. They change fairly quick so really I was able to observe a few states until it reached done state. In terms of filter/table/creds i got some notes: 
- filter only creds that I was 'issuer'
- not sure if makes sense, noticed that either holder or issuer, can't remember which one would store connectionIds: [<ids>] to pull connection/company so state of connection can be seen (e.g. verification process like verified_us...)
- render somewhere attributes like on mouse over and also manage them better, looks like some generic helpers for formatting/checking/etc

To be fair when it comes to table columns I think they could different depending on whats relevant to the end-user

src/views/credentials/index.tsx
n3op2: > Can we simplify the copy here to `No credentials found`

indeed, this happens a bit in the background and was thinkg about "New Credential" action button.

"find replace" -  a potential future prospect: generalised table based on current lists

src/views/credentials/index.tsx
n3op2: yes! we set a css class and text depending on state I think let's catch up on this

src/controllers/credentials/index.ts
```diff
@@ -24,28 +27,38 @@ export class CredentialsController extends HTMLController {
...
+        agent_connection_id: formatted[i].connectionId,
       })
-      .filter((cred) =>
-        search !== '' ? cred?.company_name.toString().toLowerCase().includes(search.toLowerCase()) : true
-      )
-
-    req.log.info('%j', combined, credentials)
-    if (search !== '') {
-      req.log.info('applying search filter... %j', combined) // do some mapping
+      formatted[i].connection = connection
```
mattdean-digicatapult, line 35: Be careful here. The Typescript type system is lying to you and `connection` might be undefined. We should probably filter out the credential in that case

src/controllers/credentials/index.ts
```diff
@@ -24,28 +27,38 @@ export class CredentialsController extends HTMLController {
...
 
+    formatted.filter((cred) => {
+      if (!cred.attributes) return false
+      const companyName = cred.attributes.map(({ name }: { name: string }) => {
+        return name === 'company_name'
+      })
+      if (search !== '') return true
+      req.log.info('checking if %s includes %s', companyName, search)
+
+      return companyName.toString().toLowerCase().includes(search.toLowerCase())
```
mattdean-digicatapult, line 46: What is the intent of `companyName` here? You are mapping over the attributes and returning an equality check so `companyName` is a `boolean[]`. You're then `toString`ing it and filtering. Did you mean to do a `find` instead of a `map`?

src/controllers/credentials/index.ts
```diff
@@ -24,28 +27,38 @@ export class CredentialsController extends HTMLController {
...
     this.setHeader('HX-Replace-Url', search ? `/credentials?search=${encodeURIComponent(search)}` : `/credentials`)
-    return this.html(this.credentialsTemplates.listPage(combined as Credential[], search))
+    return this.html(this.credentialsTemplates.listPage(formatted, search))
+  }
+
+  private format(credentials: Credential[]): Credentials {
+    return credentials.map((cred) => {
+      return {
+        ...cred,
+        attributes: cred?.credentialAttributes || [].map(({ name, value }) => ({ [name]: value })),
```
mattdean-digicatapult, line 59: I think this should be

```
attributes: (cred?.credentialAttributes || []).map(({ name, value }) => ({ [name]: value })),
```

Note that at the moment you always map over an empty array?

src/controllers/credentials/index.ts
mattdean-digicatapult, line 1: Given the issues highlighted in this review you must be missing some unit tests here

mattdean-digicatapult (changes requested):

src/controllers/credentials/index.ts
n3op2: not always, even es-lint did not trip on this, also there is a unit test that checks if it renders conn ections

src/controllers/credentials/index.ts
n3op2: Find would work.
> What is the intent of companyName here?
to make sure that credential does have a company name that is a first column of a table. and do not render ones that do not have.

mattdean-digicatapult: This is not what this code is doing

src/controllers/credentials/index.ts
n3op2: thanks, just updated to make sure that connections if defined

src/controllers/credentials/index.ts
n3op2: just updated to use find along with some additional tests

src/controllers/credentials/index.ts
n3op2: ended up not wrapping in parenthesis, since it was working I did not change it

src/controllers/credentials/index.ts
n3op2: updated

mattdean-digicatapult:
Screenshot 2024-10-09 at 10 56 50

n3op2 (commented):
LGTM

dblane-digicatapult (approved):

mattdean-digicatapult (approved):

PR #196, opened 2024-10-10
Bugfix/vr 245

n3op2:
# Pull Request

## Checklist
- [ ] I have performed a self-review of my own code.

## PR Type

- [x] Bug Fix

## Linked tickets

https://digicatapult.atlassian.net/browse/VR-245

## High level description

Fixing partial query, it was completely broken.
This will allow for normal queries along with partial to be forwarded and replied, but won't do aggregation just yet.


src/controllers/queries/index.ts
```diff
@@ -500,6 +508,81 @@ export class QueriesController extends HTMLController {
       })
     )
   }
+
+  private async submitDrpcRequest({
```
mattdean-digicatapult, line 512: Ok I think there's some confusion here as to what's going on. As far as I understand this method:

1. inserts a new `query` into the `db` and generates a new `id` for that query
2. Updates the original `query` with our local contribution.
3. Sends the `drpc` request to `charlie` using the newly inserted row
4. Adds `query_rpc` row afterwards to track that this has happened

So there's a few issues:

1. This is happening in a loop so step 2 is happening multiple times. I'd pull that out of this method
2. That should also allow you to deduplicate this functionality in `scope3CarbonConsumptionStage`
3. We need a way of linking the new query with the parent. This will require new structure in the `query` table. Maybe a `nullable` column `parent_id` with a foreign key back to the same table

Please let me know if any of this is wrong

src/controllers/queries/index.ts
```diff
@@ -352,31 +352,39 @@ export class QueriesController extends HTMLController {
...
     if (partial && partialQuery) {
       if (!partial.connectionIds || !partial.productIds || !partial.quantities) {
         throw new InvalidInputError('missing a property in the request body')
       }
       req.log.info('processing partial query %j', partial)
       const size: number = this.validatePartialQuery(partial)
-      req.log.debug('partial query has been validated %j', { partial, size })
+      req.log.debug('validate partial query and submitting DRPC request %j', { partial, size })
 
       for (let i = 0; i < size; i++) {
```
mattdean-digicatapult, line 363: Maybe consider doing this in a `Promise.all` rather than serially?

src/views/queries/queriesList.tsx
```diff
@@ -23,7 +23,7 @@ export default class QueryListTemplates {
       case 'pending_their_input':
         return <p>Sent</p>
       case 'resolved':
-        return <p>Received</p>
+        return <p>Received</p> // TODO confirm with Esther
```
Ellenn-A, line 26: is confirming this part of this PR or a subsequent one?

test/test.env
```diff
@@ -1,2 +1,2 @@
 COMPANY_PROFILE_API_KEY=API_KEY
-LOG_LEVEL=silent
```
Ellenn-A, line 2: I thought the 'silent' was a default for tests on purpose?

src/services/drpcEvents/index.ts
```diff
@@ -198,13 +199,17 @@ export default class DrpcEvents {
         this.logger.warn('Invalid queryId for drpc message %s', params.queryIdForResponse)
         return
       }
-      if (queryRow.query_response !== null) {
+
+      if (queryRow.query_response !== null && params.emissions) {
+        const total: number = parseInt(queryRow.query_response) + parseInt(params.emissions)
+        this.logger.info('updated query_response with aggregated total: %s', total)
+        params.emissions = total.toString()
+      } else if (queryRow.query_response !== null) {
```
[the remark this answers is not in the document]
n3op2: This was a little leftover from yesterday, have removed since it might not be relevant got `parent_id` column now

src/controllers/queries/index.ts
n3op2: updated to use Promise.all

src/controllers/queries/index.ts
n3op2: this has been updated along with new column

test/test.env
n3op2: yes, this should have not been included, reverted

src/views/queries/queriesList.tsx
n3op2: this is for follow up

src/controllers/queries/index.ts
```diff
@@ -352,31 +352,41 @@ export class QueriesController extends HTMLController {
...
-      req.log.debug('partial query has been validated %j', { partial, size })
 
-      for (let i = 0; i < size; i++) {
-        partialConnections.push({
-          connectionId: partial.connectionIds[i],
-          productId: partial.productIds[i],
-          quantity: parseInt(partial.quantities[i]),
+      await Promise.all(
+        new Array(size).fill({}).map((_, i) => {
+          if (partial.connectionIds && partial.productIds && partial.quantities) {
```
mattdean-digicatapult, line 364: this if statement can be removed as you already do this check on ln 356

src/controllers/queries/index.ts
```diff
@@ -500,6 +510,79 @@ export class QueriesController extends HTMLController {
       })
     )
   }
+
+  private async submitDrpcRequest({
+    queryId,
```
mattdean-digicatapult, line 515: I'd be tempted to rename this to `parent_id` for clarity.

src/controllers/queries/index.ts
```diff
@@ -500,6 +510,79 @@ export class QueriesController extends HTMLController {
       })
     )
   }
+
+  private async submitDrpcRequest({
```
mattdean-digicatapult, line 514: still need to deduplicate this logic in `scope3CarbonConsumptionStage`

mattdean-digicatapult (approved):
Three small comments to address. You can do them now if you want or in the next PR if you'd prefer

PR #201, opened 2024-10-15
Bugfix/vr 252

n3op2:
# Pull Request

## Checklist

## PR Type

- [x] Bug Fix

## Linked tickets

https://digicatapult.atlassian.net/browse/VR-245

## High level description

Basically it was a feature 'partial queries' but with no functionality. Added here some logic on how to handle partial queries along with some deduplications

Integration tests:
```sh
partial query aggregation
    with established connections: Alice -> Bob -> Charlie
      ✔ Alice sends a query request for co2 emissions to Bob
      ✔ syncs up queries and Bob sees forwarded to Charlie and from Alice
      ✔ persists a forwarded query on Charlies side
      and when Charlie responds to partial query
        ✔ updates both Bobs queries to resolved along with query responses
        ✔ also updates Alice query as resolved with a total of Bob and Charlie co2 emissions

```

also a few for DRPC and queries controller
>v A video:

https://github.com/user-attachments/assets/52193f08-3321-4f5d-a314-88ffff043226

## Detailed description

- update to quries controller
- updates to drpc for aggregation and tracking partial responses

## Operational impact

None, new feature


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "veritable-ui",
-  "version": "0.11.7",
+  "version": "0.11.8",
```
mattdean-digicatapult, line 3: I feel this deserves a minor version bump

src/controllers/queries/index.ts
```diff
@@ -511,26 +465,81 @@ export class QueriesController extends HTMLController {
...
+    return this.html(
+      this.scope3CarbonConsumptionResponseTemplates.newScope3CarbonConsumptionResponseFormPage({
+        formStage: 'success',
+        company: connection,
+        query: queryRow,
+      })
+    )
+  }
+
+  private async submitQueryRequest({
```
mattdean-digicatapult, line 522: I think we can still deduplicate this functionality from `scope3CarbonConsumptionStage` as per the previous PR's comments

src/services/drpcEvents/index.ts
```diff
@@ -241,4 +249,85 @@ export default class DrpcEvents {
...
+      this.logger.warn('DRPC has failed %j', err)
+      return this.handleError(childQuery, connection)
+    }
+    const { result, error, id: rpcId } = rpcResponse
+
+    this.logger.info('persisting query_rpc response %j', rpcResponse)
+    await this.db.insert('query_rpc', {
+      agent_rpc_id: rpcId,
+      query_id: parentQuery.id,
+      role: 'client', // am I still a client when I'm a 'responder'?
```
mattdean-digicatapult, line 304: yes this is correct. Remove the comment please though

src/services/drpcEvents/index.ts
```diff
@@ -206,15 +209,20 @@ export default class DrpcEvents {
...
-      const [query] = await this.db.update(
+      const [query]: QueryRow[] = await this.db.update(
         'query',
         { id: queryRow.id },
         { query_response: params.emissions, status: 'resolved' }
       )
 
+      if (queryRow.parent_id) {
+        await this.handleParentQuery(queryRow.parent_id, queryRow, params)
+      }
```
mattdean-digicatapult, line 220: I'm somewhat concerned by error handling here, but let's roll with it for now.

src/services/drpcEvents/index.ts
```diff
@@ -241,4 +249,85 @@ export default class DrpcEvents {
       })
     }
   }
+
+  private async handleParentQuery(id: UUID, childQuery: QueryRow, params: SubmitQueryResponseRPCParams) {
+    const [parentQuery]: QueryRow[] = await this.db.get('query', { id })
+    if (!parentQuery) {
```
mattdean-digicatapult, line 255: we should check that we haven't already respponded to the parent query and error if we have

src/services/drpcEvents/index.ts
```diff
@@ -241,4 +249,85 @@ export default class DrpcEvents {
...
+        },
+      }
+    )
+
+    const [connection] = await this.db.get('connection', { id: parentQuery.connection_id })
+    if (!connection.agent_connection_id) return this.handleError(childQuery, connection)
+
+    let rpcResponse: DrpcResponse
+    try {
+      const maybeResponse = await this.cloudagent.submitDrpcRequest(
```
mattdean-digicatapult, line 279: I think we're missing a query here to make sure that all queries that this is the parent of are in state `resolved`

mattdean-digicatapult (changes requested):
I think we need some integration tests for:

* submitting partial queries (setup a query and then test that submitting a partial query forwards a query to charlie and doesn't respond to alice)
* cascading the results (setup a query and a partial response. Respond as charlie and check the result is correctly observed by alice and bob)

You may also want to consider an e2e test for the full flow

src/services/drpcEvents/index.ts
n3op2: to be fair this might just cover the multi forwards excluding expiry date/timeout

src/services/drpcEvents/index.ts
n3op2: mmm... thought we should handle and In my mind I've a vision of simplifying errors e.g. single try/catch or `if (this) doThat()` and break down into smaller functions to avoid nesting where we do database lookups/queries but as it does change the scope and handle throws instead.

So like in scope3co2 we return error html

```
      this.scope3CarbonConsumptionTemplates.newScope3CarbonConsumptionFormPage({
        formStage: 'error',
```

but instead we log and handle:
-  db update status -> error
- if we have drpc req.id -> return something like along with setting query to state error
```
      await this.cloudagent.submitDrpcResponse(request.id, {
        error: {
          code: drpcErrorCode.INTERNAL_ERROR, // not sure
          message: <error where we have failed>,
        },
      })
```

This also brings another question: how we would like to handle (partials) when one of the childs throws an error @ the parent level

src/controllers/queries/index.ts
```diff
@@ -511,26 +465,81 @@ export class QueriesController extends HTMLController {
...
+      log.warn('DRPC has failed %j', err)
+      return this.handleError(log, queryRow, connection, undefined, err)
+    }
+    const { result, error, id: rpcId } = rpcResponse
+
+    log.info('persisting query_rpc response', rpcResponse)
+    await this.db.insert('query_rpc', {
+      agent_rpc_id: rpcId,
+      query_id: queryRow.id,
+      role: 'client', // am I still a client when I'm a 'responder'?
```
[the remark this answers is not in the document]
n3op2: no worries at all, as you can see I left it intact:) will remove

src/controllers/queries/index.ts
n3op2: This has been updated, so know we have single function (yay!) `submitDrpcRequest()` that works for partials/stage/regular

package.json
n3op2: agreed, will update in the next pushes

src/services/drpcEvents/index.ts
n3op2: removed

src/services/drpcEvents/index.ts
n3op2: this also has been included right after

docker-compose.yml
```diff
@@ -219,20 +219,20 @@ services:
         npm start | pino-colada"
     environment:
       - NODE_ENV=production
-      - LOG_LEVEL=trace
+      - LOG_LEVEL=debug
       - DB_HOST=postgres-veritable-ui-charlie
       - DB_NAME=veritable-ui
-      - API_SWAGGER_TITLE=Charlie
+      - API_SWAGGER_TITLE=Bob
```
dblane-digicatapult, line 225:
```suggestion
      - API_SWAGGER_TITLE=Charlie
```

src/controllers/queries/index.ts
```diff
@@ -359,16 +318,21 @@ export class QueriesController extends HTMLController {
...
               parentId: queryRow.id,
+              method: 'submit_query_request',
               connectionId: partial.connectionIds[i],
               log: req.log,
               localQuery: {
+                query: 'Scope 3 Carbon Consumption',
                 quantity: parseInt(partial.quantities[i]),
                 productId: partial.productIds[i],
-                emissions,
+                emissions, // TODO investigate this before asking for review, should not be needed?
```
mattdean-digicatapult, line 332: correct this should not be needed

n3op2: oops, thought I';ve removed glad left a comment, sadly, this is another leftover

src/services/drpcEvents/index.ts
```diff
@@ -241,4 +249,76 @@ export default class DrpcEvents {
...
+      const [connection] = await this.db.get('connection', { id: parentQuery.connection_id })
+      if (!connection.agent_connection_id) throw new Error('missing agent_id')
+
+      const allChilds = await this.db.get('query', { parent_id: id })
+      if (!allChilds.every(({ status }) => status === 'resolved')) {
+        this.logger.info('not all childs have responded. returning.')
+        return
+      }
+      // adding childs and parent emissions
+      const total: number = parseInt(childQuery.details.emissions) + parseInt(params.emissions || '0')
```
mattdean-digicatapult, line 274: This is only summing the current child. As you've already fetched `allChilds` you should be able to calculate from that

src/services/drpcEvents/index.ts
mattdean-digicatapult, line 1: I think there will need to be new unit tests for the partialQuery cases as well

src/controllers/queries/index.ts
n3op2: it has been removed along with property and unit tests

src/services/drpcEvents/index.ts
n3op2: indeed, thanks for flag up, since I was testing with single forward did not notice, it meant to have TODO prefix

src/services/drpcEvents/index.ts
n3op2: in terms of this I've created the below scenarios in integration test suite:

```
partial query aggregation
    with established connections: Alice -> Bob -> Charlie
      ✔ Alice sends a query request for co2 emissions to Bob
      ✔ syncs up queries and Bob sees forwarded to Charlie and from Alice
      ✔ persists a forwarded query on Charlies side
      and when Charlie responds to partial query
        ✔ updates both Bobs queries to resolved along with query responses
        ✔ also updates Alice query as resolved with a total of Bob and Charlie co2 emissions

```

will try to squeeze some for drpc as well

src/controllers/queries/index.ts
n3op2: to be fair might going to push along with drpc unit tests shortly

src/controllers/queries/__tests__/index.test.ts
```diff
@@ -135,12 +154,17 @@ describe('QueriesController', () => {
...
+        { id: 'ccaaaaaa-0000-0000-0000-d8ae0805059e' },
+        { status: 'errored' },
+      ])
 
       expect(result).to.equal('scope3_error_scope3')
     })
   })
 
-  describe('query responses', () => {
+  describe.only('query responses', () => {
```
mattdean-digicatapult, line 167: remove only

mattdean-digicatapult (changes requested):

src/controllers/queries/__tests__/index.test.ts
n3op2: Ooops, this should have been update, added some unit tests for DRPC events as well, but well, they could be better.

docker-compose.yml
```diff
@@ -155,7 +155,7 @@ services:
       - SMTP_HOST=smtp4dev
       - SMTP_PASS=password
       - SMTP_PORT=25
-      - SMTP_USER=username
+      - SMTP_USER=username # TODO should it not be bob?
```
Ellenn-A, line 158: I think this is the same for all personas as we only have one instance of the smtp4dev...but what do you think @mattdean-digicatapult ?

dblane-digicatapult: Single instance.

n3op2: will remove a comment. thanks tho

n3op2: and it has been updated along with some usefull copmments e.g. on new functions in regards to partial query

PR #209, opened 2024-10-23
Copy changes and radio input

n3op2:
# Pull Request

## Checklist

## PR Type

- [x] Feature

## Linked tickets

https://digicatapult.atlassian.net/browse/VR-269

## High level description

Copy change along with some basic layout

## Detailed description

Screenshot 2024-10-25 at 08 50 15
Screenshot 2024-10-25 at 08 49 56
Screenshot 2024-10-25 at 08 49 49
Screenshot 2024-10-25 at 08 49 45
Screenshot 2024-10-25 at 08 49 37
Screenshot 2024-10-25 at 08 49 26


Ellenn-A (commented):
`Change all instances of scope-3 or any variation of scope-3 into total-carbon-embodiment` 
was this  ^  meant to encompass all occurrences of the word `scope-3` anywhere in the repo as well (like template or css class renames etc.) or are we only worried about occurrences of scope-3 visible to the user?

rmlearney-digicatapult:
For my $0.02 I think it should be changed throughout the codebase so as not to confuse us in future

mattdean-digicatapult:
Yes I agree we should be consistent. Especially as the term `scope-3` isn't just arbitrary in this context and does actually have a real and distinct meaning

mattdean-digicatapult:
Agreed offline that we'll do the code references to `scope-3` in a separate PR

src/controllers/queries/index.ts
```diff
@@ -478,8 +478,13 @@ export class QueriesController extends HTMLController {
         await this.db.update('query', { id: query.id }, { query_response: localQuery.emissions, status: 'resolved' })
       }
 
+      const render =
+        method === 'submit_query_request'
+          ? this.scope3CarbonConsumptionResponseTemplates.newScope3CarbonConsumptionResponseFormPage
+          : this.scope3CarbonConsumptionTemplates.newScope3CarbonConsumptionFormPage
```
mattdean-digicatapult, line 484: is this fixing a bug? This feels like more than a copy change but that we're rendering something different? Also do either of these methods care about `this` binding?

n3op2: Indeed, from while was working on partials I did use incorrect newQuerySuccess one meant to be for requester and one for responder so I used the same, to be fair even text was the same (before the copy change)

mattdean-digicatapult: Ok so after discussion this does fix a bug. Because the methods are defined as `=>` functions `this` binding shouldn't be an issue

n3op2: agreed, updated to be just like: 

```js
      if (!parentId && localQuery.emissions) {
        await this.db.update('query', { id: query.id }, { query_response: localQuery.emissions, status: 'resolved' })
        return this.html(
          this.scope3CarbonConsumptionTemplates.newScope3CarbonConsumptionFormPage({
            formStage: 'success',
            company: {
              companyName: conn.company_name,
            },
          })
        )
      }
```

Ellenn-A (dismissed):
lgtm

mattdean-digicatapult (changes requested):
Changes requested to follow in comment

mattdean-digicatapult:
First issue is on the query select page. Should not be scope 3:

Screenshot 2024-10-25 at 11 06 16

mattdean-digicatapult:
Second issue is on the query response page we should not be auto-filling the radio button with `no` we ideally want a definitive user interaction. Also The text 


> Select which suppliers contributed to the carbon embodiment of this product/component. Their responses will be automatically added to your total carbon embodiment. 

Should not appear if we are not forwarding to suppliers

Screenshot 2024-10-25 at 11 08 10

n3op2:
@mattdean-digicatapult for reason can not reply to your comments.

1. When choosing a query
Screenshot 2024-10-25 at 11 25 40

2. Partial/radio button

Screenshot 2024-10-25 at 11 30 27
Screenshot 2024-10-25 at 11 30 24

mattdean-digicatapult (approved):

PR #229, opened 2024-11-06
Updating test harness and adding a new test suite for submitting a query

n3op2:
# Pull Request

## Checklist

## PR Type

- [x] Chore

## Linked tickets

https://digicatapult.atlassian.net/browse/VR-259

## High level description

Adding missing e2e test suites along with the harness

## Detailed description

- Updated checkEmail helper to take search argument
- New helpers withConnection
- Updated existing tests to assert for 'established connection'
- Test suite for Alice putting a query through
- Reducing time of e2e tests by removing double asyncs

## Operational impact

an improvement


test/e2e/helpers/registerLogIn.ts
```diff
@@ -7,6 +7,19 @@ export interface CustomBrowserContext extends BrowserContext {
   username?: string
 }
 
+export async function reset() {
```
Ellenn-A, line 10: if you have a look at the method `withCleanAliceBobEmail` right below the one you just created, you'll notice we're doing the same thing there, feel free to rename it to reset, BUT see how that one takes an url fir alice and bob? That's because it cannot be hardcoded due to e2e tests. Have a look at `connection-slice-to-bob.spect.ts` to see where we get these urls for. Then also have a look at docker compose and docker compose e2e  to see where  these envs are coming from.

test/e2e/helpers/setupConnection.ts
```diff
@@ -1,32 +1,58 @@
...
-    invite: invite,
+  // Verify and validate a new connection - holder
+  const receive = await fetchPost(`${holderUrl}/connection/new/receive-invitation`, {
+    invite: inviteBase64,
     action: 'createConnection',
   })
+  const [holderConnectionId] = (await receive.text()).match(uuidRegex) || []
+  await fetchPost(`${holderUrl}/connection/${holderConnectionId}/pin-submission`, {
+    action: 'submitPinCode',
+    pin: issuerPin,
```
Ellenn-A, line 38: mabe the naming of issuer and holder is a bit confusing here? I understand as holder you want to submit pin and invite  FROM the issuer ...but this almost suggests to me this pin is for issuer to use ... @mattdean-digicatapult ?

test/e2e/helpers/smtpEmails.ts
```diff
@@ -54,11 +57,11 @@ async function checkEmails(host: string, port: string): Promise<{ inviteEmail: E
           const results = parsedMessages.results
 
           expect(Array.isArray(results)).toBeTruthy()
-          expect(results).toHaveLength(2)
+          // expect(results).toHaveLength(2)
```
Ellenn-A, line 60: there is a method called `findNewAdminEmail` below. for usage have a look at `connection-alice-to-bob` it is for that extra admin email you get after submitting the invite. 
I don't mind if you wan tto rewrite stuff but if so also change it in the `connection-alice-to-bob` so we are consistent

test/e2e/helpers/setupConnection.ts
```diff
@@ -1,32 +1,58 @@
...
+
+import { fetchGet, fetchPost } from '../../helpers/routeHelper.js'
 import { checkEmails, extractInvite, extractPin, getHostPort } from './smtpEmails.js'
 
-export async function withConnection(smtp4devUrl: string, aliceUrl: string, bobUrl: string) {
-  // Send an invite from Alice to Bob
-  await fetchPost(`${aliceUrl}/connection/new/create-invitation`, {
+const uuidRegex = /[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-4[0-9a-fA-F]{3}-[89ab][0-9a-fA-F]{3}-[0-9a-fA-F]{12}/g
+
+export async function withConnection(issuerUrl, holderUrl) {
```
mattdean-digicatapult, line 9: why are you reducing the configurability of this? Also you don't seem to have updated it's usage

also I understand the desire to be generic beyond `alice` and `bob` but `holder` and `issuer` aren't the right terms either. What do you think of `initiatorUrl` and `receiverUrl`?

test/e2e/helpers/setupConnection.ts
mattdean-digicatapult: I think the terms are confudsing and I've made an alternative suggestion

test/e2e/helpers/setupConnection.ts
```diff
@@ -1,32 +1,58 @@
...
 
-  // Use invite and from Bob's side
-  await fetchPost(`${bobUrl}/connection/new/receive-invitation`, {
-    invite: invite,
+  // Verify and validate a new connection - holder
+  const receive = await fetchPost(`${holderUrl}/connection/new/receive-invitation`, {
+    invite: inviteBase64,
     action: 'createConnection',
   })
+  const [holderConnectionId] = (await receive.text()).match(uuidRegex) || []
```
mattdean-digicatapult, line 35: you should do some error handling here as `holderConnectionId` can be `undefined`

test/e2e/helpers/setupConnection.ts
```diff
@@ -1,32 +1,58 @@
...
+  await fetchPost(`${holderUrl}/connection/${holderConnectionId}/pin-submission`, {
+    action: 'submitPinCode',
+    pin: issuerPin,
+    stepCount: '3',
+  })
+
+  const holderEmail = await checkEmails(host, port).then(({ results }) => results[0])
```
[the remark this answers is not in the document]
mattdean-digicatapult: I think I'd rather use the search functionality of `smtp4Dev`. That way we can select specific `from` and `to` emails. See http://localhost:5001/api/index.html#/Messages/Messages_GetSummaries

test/e2e/helpers/setupConnection.ts
n3op2: agreed, was trying to pick something that we are currently using e.g. requester responder, but then it was more of query part, happy to update to invitationUrl / receive

test/e2e/helpers/smtpEmails.ts
n3op2: This is mainly to address the multiple connections e.g. in case we need to test 10 connections which is I think very possible at some point if we carry on with this. This is a temporary measure since needed to get a third email. Will check `findNewAdminEmail` and most liekly will update to use it

test/e2e/helpers/setupConnection.ts
n3op2: I've added to check that it's not undefined along with the error, same for receiverConnectionId

test/e2e/helpers/setupConnection.ts
n3op2: This has been updated to be `receiver` and `invitator`

playwright.config.ts
```diff
@@ -21,6 +21,7 @@ export default defineConfig({
   /* Shared settings for all the projects below. See https://playwright.dev/docs/api/class-testoptions. */
   use: {
     /* Collect trace when retrying the failed test. See https://playwright.dev/docs/trace-viewer */
+    headless: true,
```
Ellenn-A, line 24: so if this will run the tests without the interface maybe update the documentation? To say `if you would like to run locally with the interface update "headless" to false` or sth like that

n3op2: when I use `npm run test:e2e` it stil opens, to be fair this can be removed since it was a suggestion from playwright when test was failing. Removing now

src/services/drpcEvents/index.ts
```diff
@@ -169,7 +169,7 @@ export default class DrpcEvents {
         params = submitQueryResponseRpcParams.parse(request.params)
       } catch (err) {
         this.logger.warn('Invalid parameters received for request %s: %o', request.id, request.params)
-        this.logger.debug('Parsing error o%', err)
+        this.logger.debug('Parsing error j%', err)
```
Ellenn-A, line 172: why change this? 
I found this : 
-> Use %j if you only need a JSON-compatible output and the object is simple (without circular references).
-> Use %o when you need a detailed inspection of the object, especially in debugging scenarios.

src/controllers/queries/index.ts
```diff
@@ -406,6 +406,7 @@ export class QueriesController extends HTMLController {
         },
       })
     } catch (err) {
+      log.warn('error occured while submitting a query request', err?.toString())
```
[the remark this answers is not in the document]
n3op2: To avoid using `"as unknown as ..."` and due to JS magic it would render undefined if err isn't undefined, to be fair in TS error is unknown since it can be Error or Exception and maybe more not 100% sure here. I could do something like
```
} catch (error) {
	if (error instanceof Error) return log.warn(error.message)
	log.warn('unknown error')
```

src/services/drpcEvents/index.ts
n3op2: So I started with %o but then noticed that in pino-colada (if not mistaken) it was not rendering so since then instead of using %o I use %j also if not mistaken this could benefit searching log in tracer (to be implemented I believe) for example search by `req.id === 'some-id' && level  === '40'` across all services in this cluster, again depends on where we will forward those logs, but I think JSON is a safer option across the board

test/e2e/helpers/smtpEmails.ts
```diff
@@ -20,6 +20,10 @@ const EmailResponseSchema = z.object({
   results: z.array(EmailItemSchema),
 })
 
+// removing smtp4dev url since only one platform
+const smtp4devUrl = process.env.VERITABLE_SMTP_ADDRESS || 'http://localhost:5001'
```
Ellenn-A, line 24: so this is now both in the e2e test file and here? could we maybe just have it in one of the files and pass it from there just so it doesn't have to be in all files.

src/controllers/queries/index.ts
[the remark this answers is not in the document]
Ellenn-A: the only reason I'm asking is cuz I saw somewhere that doing `err.toString()` gets rid of the stack trace

src/services/drpcEvents/index.ts
Ellenn-A: ok

playwright.config.ts
Ellenn-A: ah ok then it's fine I guess I though it would not open when running `npm run test:e2e`

test/e2e/helpers/smtpEmails.ts
n3op2: yea, so it should be here, since this is the same URL across all nodes, did not want to do too much refactor, but should have updated, so will do now.

src/controllers/queries/index.ts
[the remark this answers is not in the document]
n3op2: yes it does, this is also going to be updated with the below one

playwright.config.ts
n3op2: however, it has been removed

test/e2e/helpers/smtpEmails.ts
n3op2: Ok, so this has been updated along with `cleanup` method that takes array of Urls and woudl call `/reset` endpoint

test/e2e/helpers/setupConnection.ts
```diff
@@ -1,32 +1,57 @@
...
+
+  await fetchPost(`${receiverUrl}/connection/${holderConnectionId}/pin-submission`, {
+    action: 'submitPinCode',
+    pin: invitatorPin,
+    stepCount: '3',
+  })
+
+  const receiverEmail = await checkEmails('admin@veritable.com')
+  expect(receiverEmail).not.toEqual(inviteEmail)
+  expect(receiverEmail).not.toEqual(adminEmail)
```
mattdean-digicatapult, line 42: I would generally avoid assertions in the test harness. Are we relying on these anywhere?

mattdean-digicatapult (commented):
This looks very clean and shouldn't interact badly with my PR. We can merge in whatever order. One question about some assertions that are present in the test harness

test/e2e/helpers/setupConnection.ts
n3op2: yes we do have in some, this is why I put as well https://github.com/digicatapult/veritable-ui/blob/main/test/e2e/helpers/smtpEmails.ts#L91-L92 going to remove since this is no longer relevant as we can now retrieve emails by to/from/subject

test/e2e/helpers/setupConnection.ts
```diff
@@ -54,4 +54,5 @@ export async function withConnection(invitatorUrl: string, receiverUrl: string)
     pin: receiverPin,
     stepCount: '2',
   })
+  await delay(2000)
```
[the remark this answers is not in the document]
n3op2: to update connection status on other node, mmm... yes, to be fair going to replace with something like polling and checking status on both sides like we have for integration test. Since I use node fetch I can not really use playwright's wait for, well I could but then in test suite we would need to add extra step to check for if pin has been verified

@mattdean-digicatapult what do you think?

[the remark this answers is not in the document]
n3op2: Just updated to await for correct status, e.g. 'Code Required' and also for 'Send Query' or return an error "timeout"

mattdean-digicatapult (approved):

PR #242, opened 2024-11-15
Shrink docker image from 530MB -> 250MB

rmlearney-digicatapult:
# Pull Request

## Checklist
- [X] New and existing unit tests pass locally with my changes.

## PR Type

- [X] Chore

## High level description

Previous docker image was 530MB in size because the multi-stage build wasn't correctly omitting dev dependencies from the final build. 

## Detailed description

`npm ci --omit-dev` in the `service` stage of the `Dockerfile` didn't correctly omit dev dependencies before.

Have amended in order to correctly omit them at final build time with `npm ci --omit=dev`.

Also removed unnecessary installation of npm at each stage because this is already bundled within the `node:` images.


Dockerfile
```diff
@@ -3,9 +3,6 @@ FROM node:current-alpine AS builder
 
 WORKDIR /veritable-ui
 
-# Install base dependencies
-RUN npm install -g npm@10.x.x
```
mattdean-digicatapult, line 7: we still want to install latest npm so please revert these

Dockerfile
```diff
@@ -32,20 +27,17 @@ ENV NODE_ENV=${NODE_ENV}
...
 
 WORKDIR /veritable-ui
 
-RUN apk add --update coreutils curl 
-RUN npm -g install npm@10.x.x
+RUN apk add --no-cache coreutils curl
 
 COPY package*.json ./
-RUN npm ci --omit-dev
+RUN npm install --omit=dev
```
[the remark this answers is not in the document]
rmlearney-digicatapult: `npm ci` will install all dev dependencies by following `package-lock` so the image will remain the same size. If you need all dependencies then this PR doesn't help

Dockerfile
rmlearney-digicatapult: Should I switch to `npm install -g npm@latest`?

rmlearney-digicatapult: `npm update -g npm` adds 30MB
`npm install -g npm@latest` adds 40MB

Dockerfile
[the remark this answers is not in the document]
rmlearney-digicatapult: Done 👍

Dockerfile
mattdean-digicatapult: Leave it as `npm install -g npm@10.x.x` as this makes sure we're using the latest available `npm` which is matched by our `package.json` `engines` section. The extra image size is worth it. 

One additional change I might ask for though is to change the base image from `current-alpine` to `lts-alpine`? That would then also match our package json `engines`

rmlearney-digicatapult: 290MB final image size

mattdean-digicatapult (approved):

PR #413, opened 2025-06-16
VR-324-integration-playwright

rmlearney-digicatapult:
# Pull Request

## Checklist

## PR Type

- [X] Bug Fix

## Linked tickets

https://digicatapult.atlassian.net/browse/VR-324

## High level description

Only build playwright in testcontainers for e2e tests

## Detailed description

1. Separate out build of playwright container from veritable-ui Dockerfile
2. Add duplicate logic in testcontainersSetup.ts to instantiate new container
3. Use new logic in globalSetup.ts for running e2e tests

## Operational impact

Accelerated test times

## Additional context

e2e tests don't run locally because we haven't addressed the bug with calling the mock Companies House API yet

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,27 +13,6 @@ RUN npm ci
 COPY . .
 RUN npm run build
 
-# Test stage
-FROM node:lts-bookworm-slim AS test
-
-WORKDIR /veritable-ui
-
-RUN npm install -g npm@11.x.x
-
-COPY package*.json ./
-COPY tsconfig.json ./
-
-RUN npm ci
-COPY . .
-RUN npm run build
-
-ARG NODE_ENV=test
-ENV NODE_ENV=${NODE_ENV}
-
-RUN npx playwright install --with-deps
-
-CMD ["npm", "run", "test:playwright"]
-
 # Service
 FROM node:lts-alpine AS service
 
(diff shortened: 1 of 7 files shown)


Dockerfile
mattdean-digicatapult, line 1: I assume this is so that we're not building the `test` container when trying to build the `service`? There's two options here:

1. we split this out to separate docker files as done here
2. we switch to using `buildkit` to do the build and explicitly state everywhere the `target` we want. That will cause unnecessary build steps to be run. If we want we can then deduplicate the logic in the build a bit better and have `test` copy in from `builder`.

Thoughts @dblane-digicatapult ?

rmlearney-digicatapult:
The `base` container is different between the two - one uses the `Dockerfile` and the other `Playwright.dockerfile`

mattdean-digicatapult:
This is then quite an inelegant solution and we can probavbly do better. One option is to embrace multi-stage builds and use a different build target fo each. That would mean moving to buildkit for the builds and I don't know how that integrates with the CI config. Another option is to pass a value of type `'Dockerfile | Playwright.dockerfile' to the method that brings up the container. Can even default it to `'Dockerfile'

rmlearney-digicatapult:
Path syntax is annoyingly slightly different for specifically passing `Dockerfile` vs another dockerfile

rmlearney-digicatapult:
Addressing this issue a different way

PR #428, opened 2025-07-04
Vr 358 connectioncleanup

rmlearney-digicatapult:
# Pull Request

## Checklist
- [X] New and existing unit tests pass locally with my changes.

## PR Type

- [X] Bug Fix
- [X] Code style update (formatting, local variables)

## Linked tickets

https://digicatapult.atlassian.net/browse/VR-359

## High level description

Change the way integration tests handle database and cloudagent cleanup in a more uniform manner

## Detailed description

NB no amount of correctly ordered cleanup solved the initial issue which still remains - I think this is due to the structure of the tests. We're not cleanly recreating the entire invitation and acceptance pathway in each test but rather probing parts of it independently.

This ticket evolved into something a little larger than originally intended.

There were 5 cleanup functions across the various helpers, all calling the same endpoints but in different ways. Different tests initialised in different but identical ways.

Creating uniform cleanup functions required passing an array pre-initialised agent and database objects into a method to be iterated over. This required refactoring of the code across the various helpers and integration tests.

The result is a more uniform instantiation of tests, condensed (and deleted) helper files, and a more useful collection of helpers and fixtures.

Keeping the tests as they are and instantiating cloudagents and databases semi-uniquely in each test and cajoling these into a cleanup method

## Operational impact

More legible and maintainable code amenable to further refactoring and creation of better shared methods for:
* connecting 2 peers (a shortcut method with direct database writes and a longer method following creation of an invitation email)
* sending and checking onboarding emails

## Additional context

Singular robust shared methods would allow for the creation of arbitrary network topologies for future test cases

We still have methods for email which are used uniquely by e2e tests which could be useful in integration tests. Some integration tests use email stubbing, some use the email server (which we instantiate for every test through `testcontainers` in any case)

The integration test for email uniquely uses a `promise` unlike any of the other tests and it could be incorporated into the connection test


test/helpers/fixtures.ts
```diff
@@ -24,23 +39,95 @@ export const charlie = {
...
+  get(name) {
+    if (name === 'PORT') {
+      return 3000
+    }
+    if (name === 'CLOUDAGENT_ADMIN_ORIGIN') {
+      return 'http://localhost:3100'
+    }
+    throw new Error('Unexpected env variable request')
+  },
+} as Env
```
jonmattgray, line 109: not used anywhere?

test/helpers/smtpEmails.ts
```diff
@@ -110,11 +104,11 @@ async function extractInvite(emailId: string): Promise<string | null> {
...
 }
 
-async function findNewAdminEmail(oldAdminEmailId: string): Promise<Email> {
+export async function findNewAdminEmail(oldAdminEmailId: string): Promise<Email> {
   return new Promise((resolve, reject) => {
     const options = {
-      hostname: host,
-      port: port,
+      hostname: 'localhost',
+      port: 5001,
```
Ellenn-A, line 111: instead of hardcoding the 'localhost' and '5001' multiple times in this file...would it be better to make them variables and then use it throughout the file? (I realise it is unlikely we'd change this anytime soon...but usually we keep stuff that is repeated multiple times in a variable so that if it ever needs changing we only need to change it in one place

test/helpers/fixtures.ts
rmlearney-digicatapult: No, but included for completeness

test/helpers/smtpEmails.ts
rmlearney-digicatapult: Agreed. I think another ticket if possible to create `test.env` and setup all of the ports that we pull into e.g. `/test/init.ts` and `/test/globalSetup.ts` from one place.

PR #436, opened 2025-07-09
Bank details query

jonmattgray:
# Pull Request

## Checklist
- [ ] Have you read Digital Catapult's [Code of Conduct](https://github.com/digicatapult/.github/blob/main/CODE_OF_CONDUCT.md)?
- [ ] I have performed a self-review of my own code.
- [ ] My changes generate no new warnings.

## PR Type

- [ ] Bug Fix
- [ ] Chore
- [ ] Feature
- [ ] Documentation Update


src/views/queries/__tests__/requestCo2embodiment.test.ts
```diff
@@ -1,82 +0,0 @@
-import { expect } from 'chai'
```
jonmattgray, line 1: I could rewrite these to work with the generic `QueryFormTemplates`, but if we're moving from snapshots anyway, seems redundant

src/controllers/queries/index.ts
```diff
@@ -403,7 +518,7 @@ export class QueriesController extends HTMLController {
     connectionId: UUID,
     parentId: UUID | null,
     params: Omit<SubmitQueryRequest['params'], 'id' | 'createdTime' | 'expiresTime'>,
-    expiresAt: Date
+    expiresTime: Date
```
rmlearney-digicatapult, line 521: porque?

jonmattgray: across the repo it's either `expiresAt` or `expiresTime`, there was more uses of `expiresTime` so I made it like that everywhere (although the `db` column is actually called `expires_at`, that's harder to change)

src/views/queries/__tests__/requestCo2embodiment.test.ts
Ellenn-A: I think that's fair to leave it out 
I assume the e2e also covers unhappy paths then?

